PR #236, opened 2023-05-04
Improve CPUID checks

easyaspi314:
- Since CLMUL is no longer used, Nehalem can now use the Westmere path.
- CPUID checks are now short circuited when it would cause an out of range CPUID.
- `__cpuidex` is now used on MSVC to ensure that the correct subleaf is used
- `__get_cpuid` is avoided since it is buggy on Clang
- The operating system is now checked for AVX and AVX512 support.
  - This includes checking macOS for AVX512 which doesn't advertise it in xgetbv and has a bug in older versions.


lemire:
> This includes checking macOS for AVX512 which doesn't advertise it in xgetbv and has a bug in older versions.

Do you have a mac with AVX-512? These are quite uncommon and I do not have access to one. We cannot test it in CI because I don't think CI will support macOS AVX-512.

easyaspi314:
> Do you have a mac with AVX-512? These are quite uncommon and I do not have access to one. We cannot test it in CI because I don't think CI will support macOS AVX-512.

No, it is a blind check but it is based on Apple's documentation and golang's research.

lemire:
> No, it is a blind check but it is based on Apple's documentation and golang's research.

Can you omit this from the PR then? I am concerned about merging code that I cannot test. This library is widely used worldwide, and I don't want to have puzzling bug reports coming back.

I would much rather wait for people who have such systems to come and ask for the support. You could provide such support in a separate PR and then we could see if people are willing to come and help test/support it.

I am not aware of a single request from macOS use for AVX-512 support from simdutf.

It is much better for us that a few users with legacy systems end up  with working AVX2 kernels, than having them with broken builds/support.

easyaspi314:
> I am not aware of a single request from macOS use for AVX-512 support from simdutf.

Before, it was using AVX-512 anyways because it was marked in CPUID. This change prevents AVX2 and AVX512 from being used on operating systems that *don't* support it.

lemire:
> Before, it was using AVX-512 anyways because it was marked in CPUID. This change prevents AVX2 and AVX512 from being used on operating systems that don't support it.

You will need to elaborate because I wrote and tested much of this code under x64 macOS. I am typing this on the last iMac that Apple with x64 processors, and it does not detect AVX-512.

If there are bugs in the current code that motivate changes, please provide sufficient details so that we can test and reproduce the bugs.

Let me restate this differently. The simdutf has a wide distribution. It has been part of Node.js for several versions now. We don't have bug reports. You are saying that our code is buggy. That's entirely possible, but if so, you should be able to allow others to verify these bugs.

I am not taking bug fixes based on faith. I need to be able to verify the bugs and verify that the code changes fix the bugs.

easyaspi314:
So to confirm:
 - Your processor supports the AVX512 instructions.
 - You are on macOS >= 12.2.
 - The new code is not correctly detecting AVX512.

To make sure I am doing the right system calls, this should output something like
```
sysctl retval: 0
output: 1
size: 8
darwin x.y
before avx512 insn: 0 or e0
after avx512 insn: e0
```

```c
#include <stdlib.h>
#include <stdio.h>
#include <sys/sysctl.h>
#include <utsname.h>
#include <stdint.h>

static inline uint64_t xgetbv() {
#if defined(_MSC_VER)
  return _xgetbv(0);
#else
  uint32_t xcr0_lo, xcr0_hi;
  asm volatile("xgetbv\n\t" : "=a" (xcr0_lo), "=d" (xcr0_hi) : "c" (0));
  return xcr0_lo | ((uint64_t)xcr0_hi << 32);
#endif
}

__attribute__((target("avx512f")))
int main()
{
  uint64_t has_avx512 = 0;
  size_t size = sizeof(uint64_t);
  long rval = sysctlbyname("hw.optional.avx512f", &has_avx512, &size, NULL, 0);
  printf("sysctl retval %ld\noutput: %lx\nsize: %lx\n", rval, has_avx512, size);
  int major = 0, minor = 0;
 
  struct utsname uname_buf = {0};
  uname(&uname_buf);
  if (sscanf(uname_buf.release, "%d.%d", &major, &minor) != 2) {
    puts("scanf fail");
  } else {
    printf("darwin %d.%d\n", major, minor);
  }

  uint64_t xcr0 = xgetbv();
  printf("before avx512 insn: %lx\n", xcr0 & (7 << 5));
  // _mm512_set1_epi8(0xFF)
  __asm__ __volatile__("vpternlogd $0xFF, %%zmm0, %%zmm0, %%zmm0" ::: "memory", "zmm0");
  xcr0 = xgetbv();
  printf("after avx512 flags: %lx\n", xcr0 & (7 << 5));
}
```

lemire:
@easyaspi314

The case that would be a concern is one where AVX-512 instructions are used when they are not supported. Your comment in the code suggests that the current code would  not detect AVX-512 even if the processor detects it. But that is fine. It would only be a concern if macOS users with AVX-512 processors were complaining, but they are very few and not complaining. AVX-512 on macOS is dead and was only a blip historically.

Now, if you are telling me that the current code enables AVX-512 on systems that do not support it, then that's a problem I care about. But I'd like to learn more about it so that we can reproduce it.

easyaspi314:
> only a blip historically

Given how this is a widely used library as you say, these historical blips should be taken care of. If I didn't care about historical blips I 

Given how I don't have a system with AVX512 I cannot test the macOS option, however this is exactly how Apple wants us to test it, and equivalent to [the workaround in golang's issue thread](https://github.com/golang/go/issues/49233).

Basically, AVX512 on macOS has some weirdness by default:

 - Because it would require > 2 KB saved per thread, the kernel does not initially enable AVX512 in XCR0.
 - The correct way of testing is via the sysctl `hw.optional.avx512f`, [as documented by Apple](https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code).
 - Upon executing an AVX512 instruction, the kernel will catch the illegal instruction, enable support (marking it in XCR0), and then resume execution as if nothing happened.
 - In macOS < 12.2, there was a bug with this shortcut where if `zmm16-zmm31` and the upper 32 bytes of `zmm0-zmm15` were zero, `k0-k7` were not properly saved on a context switch.

This both checks for AVX512 properly and makes sure that it doesn't cause issues on older versions of macOS.

However, I can say for certain that the existing AVX check was incomplete if we are going for *proper* support, and I can prove it on Ubuntu 8.04 (kernel 2.6.24, far before AVX was added in 2.6.30) on my laptop which supports AVX2. 

Here is a simplified version of the test that just tests for AVX2 using the method in simdutf before and after this PR, and tries to execute `vpcmpeqd ymm0, ymm0, ymm0`.

image

easyaspi314:
Also by the way I was able to validate that the Westmere path works and is now selected on a Nehalem box that lacks clmul. All tests pass.

EgorBo:
We've had an issue with avx-512 on macOS and afair we just disabled avx-512 support for mac cc @tannergooding

tannergooding:
MacOS supports AVX512 differently from everyone else and does it by catching a faulting instruction and dynamically enabling the support per thread.

See https://github.com/apple/darwin-xnu/blob/main/osfmk/i386/fpu.c#L174

aidarkhanov:
I ran the tests from https://github.com/easyaspi314/simdutf successfully. My laptop is i3 early 2020 MacBook Air with hardware enabled AVX-512 instructions. 

Screenshot 2023-05-05 at 22 37 57

Screenshot 2023-05-05 at 22 42 08

lemire:
Thanks @aidarkhanov 

I will thus merge this PR.

easyaspi314:
> I ran the tests from https://github.com/easyaspi314/simdutf successfully as instructed in https://twitter.com/lemire/status/1654552918417199105. My laptop is i3 early 2020 MacBook Air with hardware enabled AVX-512 instructions. If you need more information, let me know.
> 
> Screenshot 2023-05-05 at 22 37 57
> 
> Screenshot 2023-05-05 at 22 42 08

That isn't the right branch.

lemire:
@EgorBo @tannergooding Thanks for the useful feedback. Much appreciated.

PR #492, opened 2024-08-11
fix error in convert_valid_utf16le_to_latin1 fallback

pauldreik:
the implementations were inconsistent, the fallback accepted the out of range input (bad) while the others did not (good).

### DIFF
--- a/src/scalar/utf16_to_latin1/valid_utf16_to_latin1.h
+++ b/src/scalar/utf16_to_latin1/valid_utf16_to_latin1.h
@@ -15,8 +15,13 @@ inline size_t convert_valid(const char16_t* buf, size_t len, char* latin_output)
 
   while (pos < len) {
     word = !match_system(big_endian) ? utf16::swap_bytes(data[pos]) : data[pos];
-    *latin_output++ = char(word);
-    pos++;
+    if (word <= 0xFF) {
+        *latin_output++ = char(word);
+        pos++;
+    } else {
+        // can't be represented as latin1 - return error
+        return 0;
+    }
   }
 
   return latin_output - start;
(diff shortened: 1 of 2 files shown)


tests/convert_valid_utf16be_to_utf8_tests.cpp
```diff
@@ -16,6 +16,12 @@ namespace {
   constexpr int trials = 1000;
 }
 
+TEST(valid_but_out_of_range_input)
+{
+    const char16_t input = 0x2020;
+    char out;
+    ASSERT_EQUAL(0, implementation.convert_valid_utf16le_to_latin1(&input, 1, &out));
```
lemire, line 23: You are providing an input that breaks the assumption of the function.

Screenshot 2024-08-11 at 2 58 57 PM

src/scalar/utf16_to_latin1/valid_utf16_to_latin1.h
```diff
@@ -15,8 +15,13 @@ inline size_t convert_valid(const char16_t* buf, size_t len, char* latin_output)
...
   while (pos < len) {
     word = !match_system(big_endian) ? utf16::swap_bytes(data[pos]) : data[pos];
-    *latin_output++ = char(word);
-    pos++;
+    if (word <= 0xFF) {
+        *latin_output++ = char(word);
+        pos++;
+    } else {
+        // can't be represented as latin1 - return error
+        return 0;
```
lemire, line 23: Well, now you are turning a function that assumes that the input is valid latin1 into one that validates.

Why would you use that function then when you have `convert_utf16[be/le]_to_latin1`?

pauldreik:
On the question why I use this, it is because I want to exercise all code from the fuzzer. Breaking the input requirement is obviously not ok, so I will remove these from the fuzzer.

With that said, the other implementations seemed to guard from this and I thought I would fix the odd one out. But I get your message. closing!

PR #560, opened 2024-09-04
extend the misc fuzzer to cover convert_latin1_to_utf8_safe()

pauldreik:
I wonder if there is anything more than could be asserted than I already do? See in particular line 153.


lemire (approved):
This looks good. I am sure that we could make it better, but it is already reasonable.

PR #632, opened 2025-01-05
Span API followup

pauldreik:
This adds checking for concept and span support as suggested in https://github.com/simdutf/simdutf/pull/558#pullrequestreview-2530799530 .

I also took the liberty of

- move include of `<tuple>` to where it is used.
- get rid of unused macros SIMDUTF_BEGIN_DEBUG_BLOCK etc.
- remove include of `<cassert>` since it was not used
- fix minor issues in README.md


lemire (approved):
LGTM

PR #639, opened 2025-01-07
Add missing include headers for `std::pair`

atetubou:
This is to fix build error when we set use_libcxx_modules=true in [chromium build](https://logs.chromium.org/logs/chromium/buildbucket/cr-buildbucket/8726426032921605025/+/u/compile/raw_io.output_text_failure_summary_).


pauldreik:
thanks! looks good.

PR #652, opened 2025-01-14
Simplify calculating length latin1 <=> UTF16/UTF32

WojciechMula:
All these methods simply return the length of input string, as we don't check for a BOM marker.


lemire (approved):
❤️

pauldreik (approved):
nice cleanup!

PR #680, opened 2025-02-16
x86 targets: simplify UTF-32 to UTF-8

WojciechMula:
There is not point in vectorized validating for code points greater than 0x10_ffff. The blocks having code points with non-zero higher 16-bits are processed by a scalar code that detects this kind of error anyway.

### DIFF
--- a/src/haswell/avx2_convert_utf32_to_utf8.cpp
+++ b/src/haswell/avx2_convert_utf32_to_utf8.cpp
@@ -7,7 +7,6 @@ avx2_convert_utf32_to_utf8(const char32_t *buf, size_t len, char *utf8_output) {
   const __m256i v_f800 = _mm256_set1_epi16((uint16_t)0xf800);
   const __m256i v_c080 = _mm256_set1_epi16((uint16_t)0xc080);
   const __m256i v_7fffffff = _mm256_set1_epi32((uint32_t)0x7fffffff);
-  __m256i running_max = _mm256_setzero_si256();
   __m256i forbidden_bytemask = _mm256_setzero_si256();
 
   const size_t safety_margin =
@@ -17,7 +16,6 @@ avx2_convert_utf32_to_utf8(const char32_t *buf, size_t len, char *utf8_output) {
   while (end - buf >= std::ptrdiff_t(16 + safety_margin)) {
     __m256i in = _mm256_loadu_si256((__m256i *)buf);
     __m256i nextin = _mm256_loadu_si256((__m256i *)buf + 1);
-    running_max = _mm256_max_epu32(_mm256_max_epu32(in, running_max), nextin);
 
     // Pack 32-bit UTF-32 code units to 16-bit UTF-16 code units with unsigned
     // saturation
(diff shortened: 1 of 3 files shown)


pauldreik:
CI-fuzz (and oss-fuzz) usually don't run AVX512. I recommend running the fuzzers for a short while before merging (I can do it if you wish). the CI-fuzz failure seems to be real.

tests/convert_utf32_to_utf8_tests.cpp
```diff
@@ -150,4 +150,26 @@ TEST(special_cases) {
   ASSERT_EQUAL(utf8size, utf8len);
 }
 
+TEST(fuzz_7cfce3ad85893b9ca6431be21591ba8b1aa93713) {
+  const uint8_t input[] =
```
pauldreik, line 154: this must have an alignas(char32_t)

WojciechMula:
> CI-fuzz (and oss-fuzz) usually don't run AVX512. I recommend running the fuzzers for a short while before merging (I can do it if you wish). the CI-fuzz failure seems to be real.

I added a test, compiled with sanitizers, but it didn't show an issue. However, the second CI run[1] instantly reproduced the issue. Have to learn how to run fuzzing locally.

[1] https://github.com/simdutf/simdutf/actions/runs/13358198985/job/37303864709

tests/convert_utf32_to_utf8_tests.cpp
```diff
@@ -150,4 +150,26 @@ TEST(special_cases) {
...
+      "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x80"
+      "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+      "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+      "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+      "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+      "\x00'"
+      "\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+      "\xff\x00\x00\x00";
+
+  const char32_t *utf32 = reinterpret_cast<const char32_t *>(input);
```
pauldreik, line 165: I think this is only legal as of C++20 (since this paper: https://wg21.link/P0593 ) and this is supposed to compile as C++11 as well.

to avoid this language lawyer gymnastics (which I am not very good at :) it would be easier to provide the data as char32_t directly?

pauldreik (changes requested):

pauldreik:
> I added a test, compiled with sanitizers, but it didn't show an issue. However, the second CI run[1] instantly reproduced the issue. Have to learn how to run fuzzing locally.

see if you can understand the instructions at https://github.com/simdutf/simdutf/tree/master/fuzz#building-the-fuzzers-locally

WojciechMula:
> > I added a test, compiled with sanitizers, but it didn't show an issue. However, the second CI run[1] instantly reproduced the issue. Have to learn how to run fuzzing locally.
> 
> see if you can understand the instructions at https://github.com/simdutf/simdutf/tree/master/fuzz#building-the-fuzzers-locally

Thank you! Will check it.

WojciechMula:
My simplification didn't take into account that the vectorized check was in fact triggered, but only in cases when the MSB of char was set. So it wasn't a hairy bug detected by fuzzer, but a rare path in the code. I changed the code a little, to make sure that all "too-big" values will be processed by the scalar fallback.

WojciechMula:
I'm closing this issue, per the above comment. There's no clear benefit from this change.

PR #687, opened 2025-02-23
benchmarks: refactor registering benchmarks

WojciechMula:
Adding a new benchmarks was rather cumbersome (and error prone). With the new code it's required to add implementation (a method), and register it in a single place.


lemire (approved):
:heart:

PR #696, opened 2025-03-02
Introduce generic validate UTF-32 procedures

WojciechMula:
Use in Westmere and Haswell implementations.

A side-effect of #676.


lemire (approved):
Note that we have no `validate_utf32le`: the UTF-32 content is assumed to be in native order. It is a reflection of the fact that, to my knowledge, UTF-32 is never used as an interchange format, but always used internally, when it is used at all.

(You named the branch [generic-validate-utf32le](https://github.com/WojciechMula/simdutf/tree/generic-validate-utf32le))

PR #712, opened 2025-03-09
tests: enable tests for big-endian architectures

WojciechMula:
Refactor tests that deal with the UTF-16 encoding to explicitly set the desired byte order of 16-bit words. This removed a lot of places where we're manually changing endianness. Only in the case of regression tests some manual work was needed.

I checked this on my PowerPC branch. It's (almost) unrelated to the PPC64 port, so I made it a separate PR.

### DIFF
--- a/tests/convert_latin1_to_utf16be_tests.cpp
+++ b/tests/convert_latin1_to_utf16be_tests.cpp
@@ -8,7 +8,8 @@
 #include <tests/helpers/test.h>
 
 namespace {
-std::array<size_t, 7> input_size{7, 16, 12, 64, 67, 128, 256};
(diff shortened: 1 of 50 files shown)


lemire (approved):

WojciechMula:
The major change is that all helper classes for transcoding accept the expected UTF-16 strings encoding. Most required byte-swapping job is done inside the class implementations, thanks to that test code can be oblivious to machine endianness. Only in the cases when a test directly manipulates UTF-16 strings (for instance, inserting unexpected surrogate chars to trigger an error), the test has to explicitly specify what is the expected UTF-16 byte order.

Other than that, I factored a helper function to common code.

PR #713, opened 2025-03-10
base64: factor out common decoding template

WojciechMula:
This is the first step of larger refactoring. While base64 decoding itself is not complicated, the correct handling of different options we provide is quite difficult.

Having this piece of code in a single place would make future extending the code (like adding a new target) easier and less error-prone.

The main change is making most free-standing procedures methods of the `block64` class.

### DIFF
--- a/src/generic/base64.h
+++ b/src/generic/base64.h
@@ -1,0 +1,241 @@
+/**
+ * References and further reading:
+ *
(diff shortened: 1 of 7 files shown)


lemire (approved):
I think that the haswell and westmere kernels are probably easy to fuse... other kernels are more specialized and will require more work to be made general. :-)

PR #720, opened 2025-03-18
Add Olaf

WojciechMula:


lemire (approved):
Note that this list is not at all complete!

PR #770, opened 2025-04-30
accelerate atomic base64 without simd

pauldreik:
this is to get the changes in #769  into master


pauldreik:
I think I figured out what the windows build problem in CI is (it happens on linux too):
- the benchmark uses c++20
- the c++20 api is available
- simdutf is compiled in C++11, so the internal functions are not included
- we get an undefined reference when linking the benchmark

this is a bigger problem than this PR. I guess the configuration macros SIMDUTF_SPAN and SIMDUTF_ATOMIC_REF need to be sorted out at simdutf configure time, not when including the simdutf.h from the benchmark.

src/scalar/atomic_util.h
```diff
@@ -0,0 +1,105 @@
...
+  size_t offset = reinterpret_cast<std::uintptr_t>(src) % alignment;
+  if (offset) {
+    size_t to_align = std::min(len, alignment - offset);
+    memcpy_atomic_read(dst, src, to_align);
+    src += to_align;
+    dst += to_align;
+    len -= to_align;
+  }
+
+  // Process aligned 128-bit chunks
```
pauldreik, line 38: this comment is not correct

src/scalar/atomic_util.h
```diff
@@ -0,0 +1,105 @@
...
+inline void memcpy_atomic_read(char *dst, const char *src, size_t len) {
+  static_assert(std::atomic_ref<char>::required_alignment == sizeof(char),
+                "std::atomic_ref requires the same alignment as char_type");
+  // We expect all 64-bit systems to be able to read 64-bit words from an
+  // aligned memory region atomically. You might be able to do better on
+  // specific systems, e.g., x64 systems can read 128-bit words atomically.
+  constexpr size_t alignment = sizeof(uint64_t);
+
+  // Lambda for atomic byte-by-byte copy
+  auto memcpy_atomic_read = [](char *bytedst, const char *bytesrc,
```
pauldreik, line 19: this has the same name and signature as the function, which makes it a shadow and harder to understand.

src/scalar/atomic_util.h
```diff
@@ -0,0 +1,105 @@
...
+    size_t to_align = std::min(len, alignment - offset);
+    memcpy_atomic_read(dst, src, to_align);
+    src += to_align;
+    dst += to_align;
+    len -= to_align;
+  }
+
+  // Process aligned 128-bit chunks
+  while (len >= alignment) {
+    auto *dst_aligned = reinterpret_cast<uint64_t *>(dst);
```
pauldreik, line 40: I don't think it's correct to assume destination is aligned.

src/scalar/atomic_util.h
```diff
@@ -0,0 +1,105 @@
...
+  size_t offset = reinterpret_cast<std::uintptr_t>(dst) % alignment;
+  if (offset) {
+    size_t to_align = std::min(len, alignment - offset);
+    memcpy_atomic_write(dst, src, to_align);
+    dst += to_align;
+    src += to_align;
+    len -= to_align;
+  }
+
+  // Process aligned 128-bit chunks
```
pauldreik, line 85: this comment is not correct

src/scalar/atomic_util.h
pauldreik: I prefixed it with bbb_ (for byte-by-byte)

src/scalar/atomic_util.h
pauldreik: fixed this

src/scalar/atomic_util.h
pauldreik: fixed it

PR #772, opened 2025-04-30
Fix scope error in atomic thread test

pauldreik:
this fixes two minor issues.

The first one is a scope issue in the atomic thread test. I discovered this by accident when fiddling around with the threaded test when running without thread sanitizer, but with address sanitizer.

The other is a temporary hack I committed by mistake. I believe it is harmless because it is only included in c++20 anyway, but it shall be removed regardless.


lemire (approved):
Let us merge.

PR #876, opened 2025-12-13
introduce cmake option SIMDUTF_FAST_TESTS

pauldreik:
Our CI takes a long time. Part of it is because the tests take a long time. When working with the emulated aarch64 ci job, it timed out with the default 1500 seconds limit.

This PR introduces a cmake option SIMDUTF_FAST_TESTS which when set to on (default is off) adjusts the parameters of some tests to make them run with fewer iterations/smaller sizes so they finish faster. All the tests still run, just reduced. An example: the tests that use random data use fewer iterations.

Testing a release build locally on amd64, I get
 - SIMDUTF_FAST_TESTS=On: about 5 seconds test time (single threaded)
 - SIMDUTF_FAST_TESTS=Off: about 150 seconds test time (single threaded)

This also modifies ci jobs to run building and testing in parallel.

### DIFF
--- a/.github/workflows/macos-latest.yml
+++ b/.github/workflows/macos-latest.yml
@@ -18,12 +18,12 @@ jobs:
           mkdir build
           cd build
           cmake -DCMAKE_CXX_FLAGS=-Werror -DSIMDUTF_FAST_TESTS=On -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX:PATH=destination -DBUILD_SHARED_LIBS=${{matrix.shared}} ..   -DSIMDUTF_BENCHMARKS=OFF
-          cmake --build .
-          ctest --output-on-failure
+          cmake --build . --parallel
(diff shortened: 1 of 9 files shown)


pauldreik:
@lemire  this seems to work well now. total time for ci is much lower than before.

lemire:
> This also modifies ci jobs to run building and testing in parallel.

One sideeffect of increased parallelism is increased memory usage... which is why I typically err on the side of not using too much parallelism in CI.

But, clearly, it works currently, so my concern is not substantiated as of now.

lemire (approved):
Fantastic work.

pauldreik:
> > This also modifies ci jobs to run building and testing in parallel.
> 
> One sideeffect of increased parallelism is increased memory usage... which is why I typically err on the side of not using too much parallelism in CI.
> 
> But, clearly, it works currently, so my concern is not substantiated as of now.

agree!  I had the same reasoning.

PR #877, opened 2025-12-16
Treat C++20 char8_t as byte-like

leezaj:
Adds `char8_t` as a byte-like type, which in turn enables C++20 `std::u8string` and `std::u8string_view` (aliases for `std::basic_string<char8_t>` and `std::basic_string_view<char8_t>`) objects to be passed in as spans. Right now, simdutf fails to compile if users pass these objects.

[cppreference states](https://en.cppreference.com/w/cpp/language/types.html#char8_t) (emphasis mine):

>char8_t — type for UTF-8 character representation, required to be large enough to represent any UTF-8 code unit (8 bits). **It has the same size, signedness, and alignment as unsigned char (and therefore, the same size and alignment as char and signed char),** but is a distinct type. _(since C++20)_

This is enabled in >=C++20 only, and it won't affect older versions.

### DIFF
--- a/include/simdutf/implementation.h
+++ b/include/simdutf/implementation.h
@@ -52,9 +52,9 @@ namespace detail {
  * are all distinct types.
  */
 template <typename T>
-concept byte_like = std::is_same_v<T, std::byte> ||   //
-                    std::is_same_v<T, char> ||        //
-                    std::is_same_v<T, signed char> || //
+concept byte_like = std::is_same_v<T, std::byte> ||     //
+                    std::is_same_v<T, char> ||          //
+                    std::is_same_v<T, signed char> ||   //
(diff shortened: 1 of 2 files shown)


pauldreik:
Thanks for the fix! I think it looks good. I am however not sure if we should guard it with the [__cpp_lib_char8_t](https://en.cppreference.com/w/cpp/experimental/feature_test.html#cpp_lib_char8_t) feature macro?
The support seems to be pretty solid: [https://cppstat.dev/?search=char8_t](https://cppstat.dev/?search=char8_t )

I lean towards merging it as is (provided the CI tests go through).

pauldreik:
All the tests go through. Let's merge it and if there are any problems, we fix them later.

PR #878, opened 2025-12-17
Include validate_utf16le_as_ascii inside UTF16 and ASCII features

leezaj:
Right now `validate_utf16le_as_ascii` is not inside `#if SIMDUTF_FEATURE_UTF16 && SIMDUTF_FEATURE_ASCII`, even though `validate_utf16be_as_ascii` is. This change fixes that to make things more consistent.


pauldreik:
Thanks!

lemire:
We would need better tests.

pauldreik:
> We would need better tests.

Agree. There is an issue https://github.com/simdutf/simdutf/issues/855 for this.

PR #879, opened 2025-12-20
Improving the performance of validate_ascii

lemire:
Short title (summary):

The validate_ascii function is made faster for most kernels

Description

@ChALkeR first raised the issue in https://github.com/nodejs/node/pull/61122

Both validate_ascii and validate_ascii_with_errors are very fast, but when the content is pure ascii, validate_ascii should be faster. This PR fixes the issue.

Type of change
- [ ] Bug fix
- [X] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

### DIFF
--- a/benchmarks/src/benchmark.cpp
+++ b/benchmarks/src/benchmark.cpp
@@ -640,9 +640,8 @@ void Benchmark::run_validate_utf8_with_errors(
   print_summary(result, size, char_count);
 }
 
-
-void Benchmark::run_validate_ascii(const simdutf::implementation &implementation,
-                                  size_t iterations) {
+void Benchmark::run_validate_ascii(
+    const simdutf::implementation &implementation, size_t iterations) {
   const char *data = reinterpret_cast<const char *>(input_data.data());
   const size_t size = input_data.size();
   volatile bool sink{false};
--- a/benchmarks/src/benchmark.h
+++ b/benchmarks/src/benchmark.h
@@ -91,10 +91,10 @@ private:
 
 private:
   void run_validate_ascii(const simdutf::implementation &implementation,
-                         size_t iterations);
+                          size_t iterations);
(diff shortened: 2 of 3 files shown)


ChALkeR:
There are also some very weird cases I noticed (not blocking this PR)
Will check later and file a separate issue (or DM me in Slack and I'll demonstrate the weirdness, but I'm not yet sure where it comes from)

src/icelake/icelake_utf8_length_from_utf16.inl.cpp
```diff
@@ -233,7 +233,7 @@ simdutf_really_inline result icelake_utf8_length_from_utf16_with_replacement(
   }
 
   size_t overshoot = 32 - (size - pos);
-  __mmask32 remaining_mask = 0xFFFFFFFFULL << overshoot;
+  __mmask32 remaining_mask(uint32_t(0xFFFFFFFFULL << overshoot));
```
ChALkeR, line 236: is this related?

lemire: No, it is not. I am using this PR to fix an unrelated warning. Technically this is unclean, but it is merely an annoying LLVM warning (there is no bug).

lemire:
> There are also some very weird cases I noticed (not blocking this PR)


If you have found issues, my preference is that you open a public issue as we are an open community project.

But if you have security issues, I can be reached privately easily by email or you can reach @pauldreik.

ChALkeR:
Atm it's just in the "timing seems very strange and needs investigation" stage, filing an issue with that is too early - it lacks data for that yet

It's not a security issue

Given that preference, I'll file this sometime later when I'll get more understanding on what's happening in that specific case.

PR #880, opened 2025-12-23
oss-fuzz: Add unit testing build in oss-fuzz build script

arthurscchan:
OSS-Fuzz has a new Chronos feature that allows fast rebuilding of OSS-Fuzz-integrated projects and checks whether the project code or fuzzers still pass their unit tests after patching. To enable this feature, the unit tests of integrated projects need to be built. Therefore, this PR modifies the build script used by OSS-Fuzz to also build the unit tests for this project. The OSS-Fuzz-side changes for this project's integration can be found here: https://github.com/google/oss-fuzz/pull/14647.


fuzz/build.sh
```diff
@@ -54,7 +54,7 @@ fi
 
 
 cmake -B $BUILD -S . \
-      -DSIMDUTF_TESTS=Off \
+      -DSIMDUTF_TESTS=On \
```
pauldreik: please instead use a variable ENABLE_TESTS which is set to On in the else branch on row 46 and Off otherwise.

pauldreik (changes requested):
Had not heard about chronos, looks good to enable automated tooling!

If this is made to speed things up, I also think it is important to set the newly introduced `-DSIMDUTF_FAST_TESTS=On`.

fuzz/build.sh
arthurscchan: Thanks for the suggestion. I have fixed accordingly.

fuzz/build.sh
```diff
@@ -57,6 +57,7 @@ fi
 
 cmake -B $BUILD -S . \
       -DSIMDUTF_TESTS=$ENABLE_TESTING \
+      -DSIMDUTF_FAST_TESTS=$ENBALE_TESTING \
```
pauldreik: there is a spelling error here, but that does not matter since this can be unconditionally set to "On".

arthurscchan: Thanks for pointing that out and sorry for the typo. I have fixed that now.

PR #881, opened 2025-12-24
interop with std::text_encoding

shikharish:
Related to #845

### DIFF
--- a/include/simdutf/encoding_types.h
+++ b/include/simdutf/encoding_types.h
@@ -3,6 +3,12 @@
 #include <string>
 #include "simdutf/portability.h"
 
+#if !defined(SIMDUTF_NO_STD_TEXT_ENCODING) &&                                  \
+    defined(__cpp_lib_text_encoding) && __cpp_lib_text_encoding >= 202306L
+  #define SIMDUTF_HAS_STD_TEXT_ENCODING 1
+  #include <text_encoding>
+#endif
+
 namespace simdutf {
 
 enum encoding_type {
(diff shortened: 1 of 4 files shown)


shikharish:
Tested on g++-15 on MacOS(M1). I believe `std::text_encoding` is only available in libstdc++ for now. clang-21 doesn't seem to support it.

shikharish:
@lemire Request for review.

include/simdutf/encoding_types.h
```diff
@@ -52,5 +58,121 @@ encoding_type check_bom(const char *byte, size_t length);
...
+  case std::text_encoding::id::UTF16BE:
+    return UTF16_BE;
+  case std::text_encoding::id::UTF32LE:
+    return UTF32_LE;
+  case std::text_encoding::id::UTF32BE:
+    return UTF32_BE;
+  case std::text_encoding::id::ISOLatin1:
+    return Latin1;
+  default:
+    return unspecified;
```
lemire, line 113: This looks very good to me and you probably have the right function but I have to ask whether we want to throw in some error handling.

cc @pauldreik

pauldreik: I am leaning towards returning a std::optional

include/simdutf/encoding_types.h
```diff
@@ -52,5 +58,121 @@ encoding_type check_bom(const char *byte, size_t length);
...
+    return std::text_encoding(std::text_encoding::id::UTF16BE);
+  case UTF32_LE:
+    return std::text_encoding(std::text_encoding::id::UTF32LE);
+  case UTF32_BE:
+    return std::text_encoding(std::text_encoding::id::UTF32BE);
+  case Latin1:
+    return std::text_encoding(std::text_encoding::id::ISOLatin1);
+  case unspecified:
+  default:
+    return std::text_encoding{};
```
pauldreik: I think it is more clear if this explicitly returns a std::text_encoding(std::text_encoding::id::unknown).

src/encoding_types.cpp
```diff
@@ -1,3 +1,4 @@
+#include "simdutf/encoding_types.h"
```
pauldreik: does this file need to be changed at all?

tests/text_encoding_tests.cpp
```diff
@@ -0,0 +1,102 @@
+#include "simdutf.h"
```
pauldreik, line 1: nice tests! but can't we do them all constexpr?

tests/CMakeLists.txt
```diff
@@ -387,6 +387,16 @@ target_link_libraries(detect_encodings_tests
...
+add_cpp_test(text_encoding_tests)
+target_link_libraries(text_encoding_tests
+  PUBLIC simdutf::tests::helpers
+         simdutf::tests::reference)
+         
+if(cxx_std_26 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
+  set_property(TARGET text_encoding_tests PROPERTY CXX_STANDARD 26)
+  set_property(TARGET text_encoding_tests PROPERTY CXX_STANDARD_REQUIRED OFF)
+endif()
+
```
pauldreik: One thing the unit tests are there to check is that it is possible to compile the project with that specific standard. This would unconditionally enable C++26 if the compiler supports it. So I would like to see these lines removed.

Instead, I suggest adding a specific C++26 build to the ci jobs. Can you add a new ci job which uses gcc 15?

pauldreik:
I added a PR with a C++26 build: https://github.com/simdutf/simdutf/pull/883

include/simdutf/encoding_types.h
shikharish: Won't this be simpler? It is also consistent with existing `BOM::check_bom`...

lemire:
No need to worry about the clang format error/warning.

include/simdutf/encoding_types.h
shikharish: @lemire @pauldreik Thoughts?

lemire: @shikharish I think that the 'pure' C++ way would be to return std::expected or std::optional. But I think what you have is practical. If @pauldreik feels strongly and wants to block it, that's another story... but he wrote 'I am leaning'.

So, if you feel you have the right design, just make sure that it is well documented and I think we can stick with your choice.

pauldreik: > So, if you feel you have the right design, just make sure that it is well documented and I think we can stick with your choice.

agree

tests/CMakeLists.txt
pauldreik: closing this comment, a C++26 job has been merged on the master branch.

shikharish:
Made the requested changes.

tests/CMakeLists.txt
```diff
@@ -387,6 +387,12 @@ target_link_libraries(detect_encodings_tests
   PUBLIC simdutf::tests::helpers
          simdutf::tests::reference)
 
+add_cpp_test(text_encoding_tests)
+target_link_libraries(text_encoding_tests
+  PUBLIC simdutf::tests::helpers
+         simdutf::tests::reference)
+         
+
```
pauldreik: nit: extra newline

src/encoding_types.cpp
```diff
@@ -64,4 +63,5 @@ size_t bom_byte_size(encoding_type bom) {
 }
 
 } // namespace BOM
+
```
pauldreik: nit: whitespace changes in this file that should be untouched

tests/text_encoding_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <tests/helpers/test.h>
+
+#ifdef SIMDUTF_HAS_STD_TEXT_ENCODING
+
+TEST(to_std_encoding_unspecified) {
+  constexpr auto enc = simdutf::to_std_encoding(simdutf::unspecified);
+  static_assert(enc.mib() == std::text_encoding::id::unknown,
+                "unexpected unknown encoding");
+  ASSERT_TRUE(true);
```
pauldreik: why do we need all the ASSERT_TRUE(true) in this test?

tests/text_encoding_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+#else
+TEST(text_encoding_not_available) {
+  // this test passes when c++26 not available
+  ASSERT_TRUE(true);
+}
+#endif // SIMDUTF_HAS_STD_TEXT_ENCODING
+
+int main(int argc, char *argv[]) { return simdutf::test::main(argc, argv); }
```
pauldreik: please use the TEST_MAIN macro

pauldreik:
let's merge this when the CI is green

PR #884, opened 2025-12-26
fix bug in utf8_length_from_utf16() on big endian

pauldreik:
there was a bug introduced in version 7.7.0 https://github.com/simdutf/simdutf/pull/863/files#diff-d1273109039f1fb5e518645f4a1560e70c431b5dab1f8ace04ed6852f1dc3f6aR2161-R2169 

the bug only matters when all of these conditions are met:

1. `utf8_length_from_utf16be_with_replacement() `would give another answer than `utf8_length_from_utf16be()` (requires invalid input)
2. the host system is big endian

the api guarantee for utf8_length_from_utf16be() is: (quoting from the documentation)

>    This function does not validate the input. It is acceptable to pass invalid UTF-16 strings but in such cases the result is implementation defined.

this is strictly speaking not wrong.

the test added in this PR fails when running on s390x (big endian), but succeeds with the bug fix applied.


lemire:
Thanks for catching this.

PR #887, opened 2025-12-30
add simdutf::binary_length_from_base64

anonrig:
Fixes https://github.com/simdutf/simdutf/issues/867

### DIFF
--- a/include/simdutf/scalar/base64.h
+++ b/include/simdutf/scalar/base64.h
@@ -703,7 +703,7 @@ simdutf_warn_unused simdutf_constexpr23 size_t
 binary_length_from_base64(const char_type *input, size_t length) noexcept {
   // Count non-whitespace characters (c > ' ') with loop unrolling
   size_t count = 0;
-  for (size_t i; i < length; i++) {
+  for (size_t i = 0; i < length; i++) {
     count += (input[i] > ' ');
   }
 


README.md
```diff
@@ -1980,6 +1980,21 @@ if(r.error) {
 }
 ```
 
+If you want to know the exact output size without having to resize afterward, you can use
```
lemire: I'd like a qualification. Something like this. « For most applications, it is preferable to compute the maximum possible binary size rather than scanning the base64 input twice—once to calculate the exact output length and again to perform the decoding. In practice, large base64 inputs rarely contain more than a small percentage of ignorable characters (such as whitespace or line breaks), making the potential overallocation negligible compared to the cost of an extra pass over the data. [Be mindful that memory allocators do not precisely allocates the memory you requested](https://lemire.me/blog/2025/12/30/by-how-much-does-your-memory-allocator-overallocates/), they pad it up, sometimes by 10% or more. Hence, it is possible for the extra work to save no actual memory in the end. »

src/scalar/base64.h
```diff
@@ -714,6 +714,38 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(
...
+      (options & base64_default_or_url)
+          ? tables::base64::to_base64_default_or_url_value
+          : ((options & base64_url) ? tables::base64::to_base64_url_value
+                                    : tables::base64::to_base64_value);
+  // Count actual base64 characters (not ignorable, not padding)
+  size_t base64_count = 0;
+  for (size_t i = 0; i < length; i++) {
+    char_type c = input[i];
+    // Check if this is a valid base64 character (code <= 63)
+    if (is_eight_byte(c) && to_base64[uint8_t(c)] <= 63) {
```
erikcorry: I was thinking more of something like:

if (c > ' ')

which would be easier to parallelize.

Only the last few characters should be checked for '='.

We never promised to filter out non-ASCII whitespace, even 0xa0 which is non-breaking space.

The scalar version should return the same as the SIMD versions, so it should not try to be clever.

README.md
```diff
@@ -2246,6 +2261,46 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(const char * input,
  */
 simdutf_warn_unused size_t maximal_binary_length_from_base64(const char16_t * input, size_t length) noexcept;
 
+/**
+ * Provide the exact binary length in bytes given the base64 input.
```
erikcorry: Let's not overpromise here, making an API that is harder to accelerate.

The original issue only promised the correct answer for well-formed base64 with optional ASCII spaces. It doesn't make any promises about only counting actual base64 characters.

lemire:
@anonrig I like @erikcorry's comment here tremendously. But we should then not take an option parameter because, among the possible options, is that we ignore everythings (as Node.js's atob does). Further, it is clear form what @erikcorry is describing that we do no validation of any kind... 

So we would just say something like 'we return the exact output size assuming that the base64 input is valid [obviously] and ignoring ASCII white-space characters.'


With Erik's simplified approach, we can even implement accelerated functions in the various kernels as needed, and it becomes a SIMD programming 101 problem (beat the autovectorizer). 

(Note that the exact function can be done sanely within our kernels without a PhD in SIMD programming... but it is going to involve non trivial arithmetic... we already have the code... but it is still going to be slower than what Erik is proposing we implement.)

lemire:
This PR does not have to include my comment above about how this is often going to be a suboptimal idea. But I will want to throw it in later because I think we need have people be aware of the tradeoff. (This function here is fine as long as the engineer is aware of the tradeoff.)

README.md
anonrig: The point of introducing this particular API is: 

- Maximal length method assumes that the underlying data structure that stores the input for the base64 decoded string is a resizable array/buffer.
- Not all environments and API structures support resizable array buffers - such as in this case, v8. ArrayBuffer need to be an ResizableArrayBuffer using that means the return value is changed. 

With these limitations, we have 2 paths in front of us:

1. Iterate the input twice, 1 for getting the exact size, and 1 for converting the base64 to binary.
2. Use maximal length method on a growable data structure on heap (like std::vector) and copy only the first N characters that are needed to the actual non-resizable data structure (which is what the current v8 does)

lemire: @anonrig That's fine but consider @erikcorry's proposal which, I think, also solves your practical problem.

anonrig: Yes I'll address it now

lemire:
@anonrig Warning: Paul (@pauldreik) turned everything constexpr so this may interact badly with this PR.

src/scalar/base64.h
anonrig: Ah I missed that part. Nice catch. Updating now...

anonrig:
@lemire No worries. I'll rebase and address the reviews and ready the PR for the next round of reviews.

README.md
```diff
@@ -1982,6 +1982,21 @@ if(r.error) {
...
 ```
 
+If you have well-formed base64 input with ASCII spaces and want to know the output size
+without having to resize afterward, you can use `binary_length_from_base64` which counts
+non-whitespace characters (excluding padding):
+
+```cpp
+std::vector<char> buffer(simdutf::binary_length_from_base64(base64.data(), base64.size()));
+simdutf::result r = simdutf::base64_to_binary(base64.data(), base64.size(), buffer.data());
+if(r.error) {
```
erikcorry: Check explicitly against SUCCESS, not with implicit bool conversion. Also missing space after 'if'.

README.md
```diff
@@ -1982,6 +1982,21 @@ if(r.error) {
 }
 ```
 
+If you have well-formed base64 input with ASCII spaces and want to know the output size
+without having to resize afterward, you can use `binary_length_from_base64` which counts
+non-whitespace characters (excluding padding):
```
erikcorry:
```suggestion
You can calculate the exact output space needed by using
`binary_length_from_base64` which produces an exact number of output
bytes if the input is well-formed. Well-formed means it contains
only valid base64 and ASCII whitespace. Invalid input can be given to
`binary_length_from_base64`. It will not detect invalid input, but the
result can be safely used to size the output buffer for `base64_to_binary`,
which does detect invalid input.
```

README.md
```diff
@@ -2248,6 +2263,34 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(const char * input,
  */
 simdutf_warn_unused size_t maximal_binary_length_from_base64(const char16_t * input, size_t length) noexcept;
 
+/**
+ * Compute the binary length from a base64 input with ASCII spaces.
+ * This function is useful for well-formed base64 inputs that may contain
+ * ASCII spaces (such as line breaks). For such inputs, the result is exact.
```
erikcorry:
```suggestion
 * This function is useful for base64 inputs that may contain ASCII spaces
 * (such as line breaks). For such inputs, the result is exact, and for any
 * inputs the result can be used to size the output buffer passed to
 * `base64_to_binary`.
```

README.md
```diff
@@ -2248,6 +2263,34 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(const char * input,
  */
 simdutf_warn_unused size_t maximal_binary_length_from_base64(const char16_t * input, size_t length) noexcept;
 
+/**
+ * Compute the binary length from a base64 input with ASCII spaces.
```
erikcorry:
```suggestion
 * Compute the binary length from a base64 input.
```

README.md
```diff
@@ -2248,6 +2263,34 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(const char * input,
...
+ * subtracts padding characters ('=') found at the end.
+ *
+ * @param input         the base64 input to process
+ * @param length        the length of the base64 input in bytes
+ * @return number of binary bytes
+ */
+simdutf_warn_unused size_t binary_length_from_base64(const char * input, size_t length) noexcept;
+
+/**
+ * Compute the binary length from a base64 input with ASCII spaces.
```
erikcorry: Same changes as above.

README.md
```diff
@@ -2248,6 +2263,34 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(const char * input,
  */
 simdutf_warn_unused size_t maximal_binary_length_from_base64(const char16_t * input, size_t length) noexcept;
 
+/**
+ * Compute the binary length from a base64 input with ASCII spaces.
+ * This function is useful for well-formed base64 inputs that may contain
+ * ASCII spaces (such as line breaks). For such inputs, the result is exact.
+ *
+ * The function counts non-whitespace characters (ASCII value > 0x20) and
+ * subtracts padding characters ('=') found at the end.
```
erikcorry:
```suggestion
 * The function ignores whitespace and does not require padding characters ('=').
```

README.md
```diff
@@ -2248,6 +2263,34 @@ simdutf_warn_unused size_t maximal_binary_length_from_base64(const char * input,
...
+ *
+ * @param input         the base64 input to process, in ASCII stored as 16-bit units
+ * @param length        the length of the base64 input in 16-bit units
+ * @return number of binary bytes
+ */
+simdutf_warn_unused size_t binary_length_from_base64(const char16_t * input, size_t length) noexcept;
```
erikcorry, line 2285: Inconsistent spacing around asterisks - we should use the 'house style' of simdutf.

include/simdutf/implementation.h
```diff
@@ -4161,6 +4161,50 @@ maximal_binary_length_from_base64(std::span<const char16_t> input) noexcept {
 }
   #endif // SIMDUTF_SPAN
 
+/**
+ * Compute the binary length from a base64 input with ASCII spaces.
+ * This function is useful for well-formed base64 inputs that may contain
+ * ASCII spaces (such as line breaks). For such inputs, the result is exact.
```
erikcorry:
```suggestion
 * Compute the binary output length from a base64 input with optional '='
 * padding.
 * For well-formed base64 inputs that may contain ASCII spaces (such as line
 * breaks) the result is exact.```

File: include/simdutf/implementation.h [hunk 12]
```diff
@@ -4161,6 +4161,50 @@ maximal_binary_length_from_base64(std::span<const char16_t> input) noexcept {
...
+  #if SIMDUTF_SPAN
+simdutf_really_inline simdutf_warn_unused size_t binary_length_from_base64(
+    const detail::input_span_of_byte_like auto &input) noexcept {
+  return binary_length_from_base64(reinterpret_cast<const char *>(input.data()),
+                                   input.size());
+}
+  #endif // SIMDUTF_SPAN
+
+/**
+ * Compute the binary length from a base64 input with ASCII spaces.
```
erikcorry: As above.

include/simdutf/implementation.h
```diff
@@ -6357,6 +6401,37 @@ class implementation {
   simdutf_warn_unused size_t maximal_binary_length_from_base64(
       const char16_t *input, size_t length) const noexcept;
 
+  /**
+   * Compute the binary length from a base64 input with ASCII spaces.
```
erikcorry, line 6426: As above

include/simdutf/implementation.h
```diff
@@ -6357,6 +6401,37 @@ class implementation {
...
+   *
+   * @param input         the base64 input to process
+   * @param length        the length of the base64 input in bytes
+   * @return number of binary bytes
+   */
+  simdutf_warn_unused size_t
+  binary_length_from_base64(const char *input, size_t length) const noexcept;
+
+  /**
+   * Compute the binary length from a base64 input with ASCII spaces.
```
erikcorry, line 6441: As above

include/simdutf/scalar/base64.h
```diff
@@ -694,6 +694,58 @@ maximal_binary_length_from_base64(InputPtr input, size_t length) noexcept {
...
+    pos--;
+    // Skip whitespace between padding characters
+    while (pos > 0 && input[pos - 1] <= ' ') {
+      pos--;
+    }
+    // Check for second '='
+    if (pos > 0 && input[pos - 1] == '=') {
+      padding++;
+    }
+  }
```
erikcorry:
```suggestion
  while (pos > 0 && padding < 2) {
    char_type c = input[--pos];
    if (c == '=') {
      padding++;
    } else if (c > ' ') {
      break;
    }
  }
```

include/simdutf/scalar/base64.h
```diff
@@ -694,6 +694,58 @@ maximal_binary_length_from_base64(InputPtr input, size_t length) noexcept {
...
+// and counting non-whitespace characters (excluding padding characters).
+// We use a simple check (c > ' ') which is easy to parallelize and matches
+// SIMD behavior. Only the last few characters are checked for padding '='.
+template <class char_type>
+simdutf_warn_unused size_t binary_length_from_base64(const char_type *input,
+                                                     size_t length) noexcept {
+  // Count non-whitespace characters (c > ' ') with loop unrolling
+  size_t count = 0;
+  size_t i = 0;
+  for (; i + 4 <= length; i += 4) {
```
erikcorry: This hand-unrolling actually prevents gcc from autovectorizing with -O3 -march=haswell. The 'dumb' version generates better code.

include/simdutf/scalar/base64.h
erikcorry: This is untested, but I think something like this is more elegant.

README.md
erikcorry: It is 'useful' for non-well-formed inputs too, although the conversion function will return error.

tests/base64_tests.cpp
```diff
@@ -3494,6 +3494,119 @@ TEST(binary_to_base64_with_lines_span_api_char) {
 
 #endif
 
+// Tests for binary_length_from_base64
```
erikcorry, line 3497: Some of the tests should take invalid input if only to verify that there is no crash or buffer overflow. Eg. Something that only consists of '=' and spaces.

include/simdutf/scalar/base64.h
lemire: I think that @erikcorry's version here works.

include/simdutf/scalar/base64.h
lemire: @erikcorry Can you review the following:

https://godbolt.org/z/4E4frcxh7

I am compiling @anonrig's code on Godbolt with -O3 and both LLVM and GCC appear to be happy to autovectorize (`-O3`).

include/simdutf/scalar/base64.h
```diff
@@ -694,6 +694,58 @@ maximal_binary_length_from_base64(InputPtr input, size_t length) noexcept {
...
+  }
+  size_t base64_count = count - padding;
+
+  // Calculate binary length from the number of base64 characters
+  // Every 4 base64 characters encode 3 binary bytes
+  // Remainder of 2 encodes 1 byte, remainder of 3 encodes 2 bytes
+  if (base64_count % 4 <= 1) {
+    return base64_count / 4 * 3;
+  }
+  return base64_count / 4 * 3 + (base64_count % 4) - 1;
```
erikcorry:
```suggestion
  return ((count - padding) * 3) / 4;
```

include/simdutf/scalar/base64.h
erikcorry: Strangely gcc fails to vectorize when using C++ as the language instead of C.
https://godbolt.org/z/aG9WWef18

The vectorization looks pretty bad though, so probably this doesn't matter.

tests/base64_tests.cpp
```diff
@@ -3494,6 +3494,119 @@ TEST(binary_to_base64_with_lines_span_api_char) {
...
+}
+
+TEST(binary_length_from_base64_char16) {
+  // Test with char16_t input
+  std::u16string input16 = u"YWJj"; // "abc"
+  ASSERT_EQUAL(
+      simdutf::binary_length_from_base64(input16.data(), input16.size()), 3);
+
+  std::u16string input16_spaces = u"  Y  Q  =  =  ";
+  ASSERT_EQUAL(simdutf::binary_length_from_base64(input16_spaces.data(),
```
erikcorry, line 3567: These inputs are so short I don't think they are testing an actual SIMD implementation, since they drop right through to the scalar cleanup. At least I wrote a broken version of the 16 bit version in another PR and the tests passed.

lemire:
I'll mark this PR as approved. We could merge, but before turning it into a release, I think we want to do more performance work. I don't have time right now, but later...

---

Performance numbers reproduced from my other PR.

---

To test this out, do the following:

```
cmake -B build -D SIMDUTF_BENCHMARKS=ON -D CMAKE_BUILD_TYPE=Release
cmake --build build --target benchmark_base64
```

Use a test file. Any would not, but you can create one like so:

```
 base64 -i ./README.md -b 76 > test.base64
```

Next run the benchmarks, first the decoding benchmark:

```
./build/benchmarks/base64/benchmark_base64 -d test.base64 -f simdutf
```

Here `-f simdutf` applies a filter so we only run the simdutf functions.

Then benchmark the length functions:

```
 ./build/benchmarks/base64/benchmark_base64 -L test.base64
```


Here is what I get on my macbook...

```
./build/benchmarks/base64/benchmark_base64 -d test.base64 -f simdutf

# current system detected as arm64.
# loading files: .
# volume: 182408 bytes
# max length: 182408 bytes
# number of inputs: 1
# decode
# the base64 data contains spaces, so we cannot use straight libbase64::base64_decode directly
simdutf::arm64                                :  15.82 GB/s  7.27 % 
simdutf::arm64 (accept garbage)               :  13.77 GB/s  6.65 % 

 ./build/benchmarks/base64/benchmark_base64 -L test.base64         

# current system detected as arm64.
# loading files: .
# volume: 182409 bytes
# max length: 182409 bytes
# number of inputs: 1
# lengths
# Benchmark only simdutf length functions (maximal and exact)
simdutf::arm64_maximal_binary_length_from_base64 :  10838.88 GB/s  inf % 
simdutf::arm64_binary_length_from_base64      :   8.31 GB/s  3.06 % 
```


So you see here that `maximal_binary_length_from_base64` is effectively free, while `binary_length_from_base64` is not.

Suppose you combine the decoding function with the maximal function... then we get `15.82 GB/s` (unchanged).

But if you combine it with the new function you get 1/(1/8.31 + 1/15.82) or `5.5 GB/s`. That is, we reduce by a factor of three the speed. It is not a small effect.

pauldreik:
@anonrig before we merge this, I think there should be a span overload and it should be constexpr. ping me if you want help or me to do it. it should be easy.

anonrig:
> @anonrig before we merge this, I think there should be a span overload and it should be constexpr. ping me if you want help or me to do it. it should be easy.

@pauldreik Thanks! I'll add it now.

@erikcorry ~~Would you mind if I follow up on your remaining recommendations after landing this PR?~~ Actually, I'll address them now.

include/simdutf/implementation.h
```diff
@@ -4176,10 +4176,19 @@ maximal_binary_length_from_base64(std::span<const char16_t> input) noexcept {
...
   #if SIMDUTF_SPAN
-simdutf_really_inline simdutf_warn_unused size_t binary_length_from_base64(
+simdutf_really_inline simdutf_warn_unused simdutf_constexpr23 size_t
+binary_length_from_base64(
     const detail::input_span_of_byte_like auto &input) noexcept {
-  return binary_length_from_base64(reinterpret_cast<const char *>(input.data()),
-                                   input.size());
+    #if SIMDUTF_CPLUSPLUS23
+  if consteval {
+    return scalar::base64::binary_length_from_base64(
```
pauldreik: this is incompatible with the signature in scalar/base64.h

I suggest you skip the constexpr_cast_ptr and just pass input.data() directly.

anonrig: Ah yes. I'll fix it now. Thanks for the review.

pauldreik:
good, please also add constexpr tests for the utf16 variant!

anonrig:
Sorry for the late response. I've added the last remaining tests, rebased and force pushed. I think this pull-request is good to go. @lemire would you mind landing this pull-request if you think it's good to go (and when tests succeed)?

lemire:
@anonrig I am happy to approve this. Obviously, it will require more work to get the optimized performance, but it is fine to do it as follow up PRs. I see that @pauldreik has a 'requested changes' status and I think that @erikcorry had comments, so I will approve but give time to these fine folks to react before we merge for good.

Merging early next week if nobody has anything to add ?

include/simdutf/implementation.h
```diff
@@ -4161,6 +4161,71 @@ maximal_binary_length_from_base64(std::span<const char16_t> input) noexcept {
 }
   #endif // SIMDUTF_SPAN
 
+/**
+ * Compute the binary length from a base64 input.
+ * This function is useful for base64 inputs that may contain ASCII spaces
```
pauldreik: spaces -> whitespace ?

if you agree, the same thing applies multiple times in the doxygen comments in this PR.

include/simdutf/scalar/base64.h
```diff
@@ -694,6 +694,42 @@ maximal_binary_length_from_base64(InputPtr input, size_t length) noexcept {
...
+binary_length_from_base64(const char_type *input, size_t length) noexcept {
+  // Count non-whitespace characters (c > ' ') with loop unrolling
+  size_t count = 0;
+  size_t i = 0;
+  for (; i + 4 <= length; i += 4) {
+    count += (input[i + 0] > ' ');
+    count += (input[i + 1] > ' ');
+    count += (input[i + 2] > ' ');
+    count += (input[i + 3] > ' ');
+  }
```
pauldreik: measuring with the provided benchmark in this PR on a file created with `head -c 1000000 /dev/random |base64 >/tmp/a`:
when I removed these loop unrolling lines, gcc increased from 6 to 9 GB/s. on clang, I got 3 GB/s regardless.

it seems like the loop unrolling at least on my x86 system is a pessimization, and I suggest removing it.

tests/base64_tests.cpp
pauldreik: I agree but I think that can be added later, if and when vectorization is implemented.

README.md
pauldreik: I agree.

But I also think this is a general problem not belonging to this PR - the documentation is duplicated. Do we need the function signatures in the README ? Or can they be auto generated/validated.

lemire: We definitely can do better as far as the documentation goes.

anonrig:
Can we merge this? Sorry I haven't been super active with this pull-request.

pauldreik:
> Can we merge this? Sorry I haven't been super active with this pull-request.

yes!

Issue #886, opened 2025-12-30
the scripts/check_feature_macros.py needs fixing


pauldreik:
the scripts/check_feature_macros.py does not seem to work as intended:

## problem 1: it is not running

it does not invoke the main() function. I pushed a change with the following diff:
```diff
Author: Paul Dreik <github@pauldreik.se>
Date:   Tue Dec 30 10:54:08 2025 +0100

    make feature macro script runnable

diff --git a/scripts/check_feature_macros.py b/scripts/check_feature_macros.py
old mode 100644
new mode 100755
index 1462a6b7..5c8ba718
--- a/scripts/check_feature_macros.py
+++ b/scripts/check_feature_macros.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import subprocess
 import re
 import os
@@ -123,4 +124,7 @@ def main():
         sys.exit(1)  # Exit with error code if errors are found
     else:
         print("All #if...#endif pairs for SIMDUTF feature macros are correctly matched with valid comments.")
-        sys.exit(0)  # Exit with success code
\ No newline at end of file
+        sys.exit(0)  # Exit with success code
+
+if __name__ == "__main__":
+    main()
```

the ci job started failing for visual studio - this script is run by the check_feature_macro test, which now seems to start running. it must have been not running before (proof in the next point).

## problem 2: it does not handle multiple lines

since a while back, the feature macros check sometimes end up on multiple lines. the script does not understand that.

the reported error is:

```plain
ERROR: include/simdutf/implementation.h:2974: #endif comment 'SIMDUTF_FEATURE_BASE64 || SIMDUTF_FEATURE_UTF16 ||' does not match #if condition 'SIMDUTF_FEATURE_BASE64 || SIMDUTF_FEATURE_UTF16 ||                         \' at line 2969
ERROR: src/arm64/implementation.cpp:17: #endif comment 'SIMDUTF_FEATURE_ASCII || SIMDUTF_FEATURE_DETECT_ENCODING ||' does not match #if condition 'SIMDUTF_FEATURE_ASCII || SIMDUTF_FEATURE_DETECT_ENCODING ||                \' at line 11
[...]

PR #889, opened 2025-12-30
fixes the check_feature_macros script.

lemire:
Fixes https://github.com/simdutf/simdutf/issues/886

Obviously, I won't win the Turing award for this code. But it only needs to be good enough to spot our mistakes.

### DIFF
--- a/scripts/check_feature_macros.py
+++ b/scripts/check_feature_macros.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import subprocess
 import re
 import os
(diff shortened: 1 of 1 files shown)


pauldreik:
well it seems to work! in the sense that it gives no warnings.

what do you say about running this as a separate ci job and remove it from the unit test? I volunteer to fix it if needed.

lemire:
@pauldreik 

> well it seems to work! in the sense that it gives no warnings.

I am 90% sure that it means that things are correct.

> what do you say about running this as a separate ci job and remove it from the unit test? I volunteer to fix it if needed.

As I wrote elsewhere, I think we want *us* to catch the mistakes when *we* code before we push a commit.

PR #897, opened 2026-01-04
Add a C API

lemire:
Short title (summary):
Add a C API

Description

This PR adds a C API to simdutf. Amazingly, I cannot find a corresponding open issue.

Note that simdutf itself depends on the C++ library. 

Purpose: This should make it easy to call simdutf from Swift, Go and friends.

To keep it simpler, I am not adding the feature macros to it... I am concerned that it will add much complexity, although it can be done later.

See https://github.com/simdutf/simdutf/pull/893

cc @MahdiBM

Type of change
- [ ] Bug fix
- [ ] Optimization
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

### DIFF
--- a/tests/simdutf_c_tests.cpp
+++ b/tests/simdutf_c_tests.cpp
@@ -1,9 +1,6 @@
 #include <tests/helpers/test.h>
 #include <cstring>
-
-extern "C" {
 #include "simdutf_c.h"
-}
 
 const char *hello = "hello";
 const size_t hello_len = 5;


README.md
```diff
@@ -2689,11 +2689,31 @@ If you use this library in your research, please cite our work:
...
+
+The C API exposes functions for validation, transcoding, size estimation, `find` helpers,
+and Base64 encode/decode helpers. Results are returned using the `simdutf_result` struct
+which contains an `error_code` field and additional fields when relevant.
 
 
+We provide a simple C demo using the C wrapper at `amalgamation_demo.c`.
+It shows validating UTF-8, converting UTF-8 to UTF-16LE and back, and checking the round-trip.
+Refer to `singleheader/README.md` for instructions. Note that the simdutf library requires
+a C++ library.
```
pauldreik: compiler?

include/simdutf_c.h
```diff
@@ -0,0 +1,330 @@
+#ifndef SIMDUTF_C_H
+#define SIMDUTF_C_H
+
+#include <stddef.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __has_include
```
pauldreik, line 15: this seems to be a C23 feature, so I think it must be changed or we require C23.
https://en.cppreference.com/w/c/preprocessor/include

src/simdutf_c.cpp
```diff
@@ -0,0 +1,576 @@
...
+  out.error = static_cast<simdutf_error_code>(r.error);
+  out.count = r.count;
+  return out;
+}
+
+/* The C wrapper depends on the library features. Only expose the C API
+   when all relevant feature is enabled. This helps the
+   single-header generator to omit the C wrapper when features are
+   disabled. */
+// clang-format off
```
pauldreik, line 15: I think we should turn on clang format again, after the next line. (I assume you wanted the preprocessor macro on one  line and that was the reason for turning off formatting)

src/simdutf_c.cpp
```diff
@@ -0,0 +1,576 @@
...
+      static_cast<simdutf::last_chunk_handling_options>(last_chunk_options),
+      decode_up_to_bad_char);
+  if (outlen)
+    *outlen = local_out;
+  return to_c_result(r);
+}
+
+} // extern "C"
+
+#endif // SIMDUTF_FEATURE_UTF8 && SIMDUTF_FEATURE_UTF16 && SIMDUTF_FEATURE_UTF32 && SIMDUTF_FEATURE_LATIN1 && SIMDUTF_FEATURE_ASCII && SIMDUTF_FEATURE_BASE64 && SIMDUTF_FEATURE_DETECT_ENCODING
```
pauldreik, line 576: see earlier comment about clang format off

tests/CMakeLists.txt
```diff
@@ -421,6 +421,12 @@ add_cpp_test(span_tests)
 target_link_libraries(span_tests
   PUBLIC simdutf::tests::helpers
          simdutf::tests::reference)
+
+# C API tests
```
pauldreik, line 425: it feels good but insufficient to have c tests that compile with cpp. one thing we want to test is that it is actually usable from c.

tests/simdutf_c_tests.cpp
```diff
@@ -0,0 +1,127 @@
+#include <tests/helpers/test.h>
+#include <cstring>
+
+extern "C" {
```
pauldreik: I think this is misplaced - the included file should (and already does) protect itself from being compiled with a C++ compiler. suggestion: remove the extern here.

README.md
lemire: I meant here at runtime. You either need to statically link a C++ library *or* link against a C++ dynamic library.

So it is less a trivial statement than "you need a C++ compiler to compile a C++ library", but really the fact that even if the library has been compiled, you still need to link against a C++ standard library. For context, see https://github.com/redis/redis/pull/14661

I'll replace 'library' by 'standard library'. Let me know it is unclear..

include/simdutf_c.h
lemire: > this seems to be a C23 feature, so I think it must be changed or we require C23.

Formally, you are correct, but both clang and gcc have supported `__has_include` for a really long time.

https://godbolt.org/z/895hcr7ff

The issue is that `uchar.h`is C11 but Apple does not have the header. I don't know why.

src/simdutf_c.cpp
lemire: done.

src/simdutf_c.cpp
lemire: done

tests/simdutf_c_tests.cpp
lemire: I don't think so. It is a C++ file!!!!!

tests/CMakeLists.txt
lemire: I have now added a test in pure C.

README.md
lemire: Description added.

lemire:
@pauldreik Making as 'ready for review'.

tests/simdutf_c_tests.cpp
pauldreik: maybe we misunderstand each other? 
the "simdutf_c.h" is a c header file, which already tolerates being compiled as c++ thanks to it's checks for the __cplusplus macro. so the extern C on row 4 here, is not needed and also makes anything (such as includes of other files) be wrapped in extern C.

so, I think row 4 and 6 should be deleted. agree?

README.md
```diff
@@ -2689,11 +2689,31 @@ If you use this library in your research, please cite our work:
 ```
 
 
-Stars
--------
 
+C wrapper (C11 or better)
```
pauldreik, line 2693: should we mark this as experimental for the next release? that way we can fix mistakes and get feedback.

pauldreik (commented):
I tested the amalgamate script, it failed with:

`FileNotFoundError: [Errno 2] No such file or directory: '/home/pauldreik/code/delaktig/simdutf/singleheader/simdutf_c.h'`

pauldreik:
@lemire  I took the liberty to push some clang format changes and ci job, hope you don't mind.

README.md
pauldreik: thanks, better!

lemire:
@pauldreik I don't mind if you push stuff, but the rvv CI changes seem to fail.

lemire:
> so, I think row 4 and 6 should be deleted. agree?

Maybe I misunderstand you. Here is what I think. If I write a function definition in C++ and I want it to be callable from C, then I put it inside an `extern 'C'` scope.

lemire:
@pauldreik I pushed two changes.

1. Fixed a logic error in the amalgamation script (thanks)
2. Marked the C API as experimental (meaning we warn about possible bugs)

pauldreik:
> @pauldreik I don't mind if you push stuff, but the rvv CI changes seem to fail.

it was not my push that caused the failure, I pushed a fix for one of the three failing riscv jobs. I have now pushed fixes for the remaining two.

pauldreik:
> > so, I think row 4 and 6 should be deleted. agree?
> 
> Maybe I misunderstand you. Here is what I think. If I write a function definition in C++ and I want it to be callable from C, then I put it inside an `extern 'C'` scope.

It is the declaration that needs to be in extern C. But that is not what I am after here, it is that the wrapping of extern C around the include is at best redundant and at worst wrong (since it unconditionally puts everything, including not functions, from the header file in extern C)

tests/simdutf_c_tests.cpp
lemire: Removed.

lemire:
@pauldreik I did not see that extern C at first, in the test file. Sorry. I just removed it.

It did not make sense, I agree.

pauldreik:
no problem! are we good to merge then?

lemire:
Merged !

PR #898, opened 2026-01-05
Base64 length for Haswell and Icelake.

erikcorry:
binary_length_from_base64 for Haswell.

Description

This is a stab at an implementation for Haswell.

It's a bit unusual in that it doesn't have a scalar clean-up loop at the end.  Instead, it reads too far and then subtracts the extra bytes again. Because all reads are aligned it can't fault.

(Sadly there is then a scalar search for '=' padding afterwards.)

Unfortunately , this sort of implementation is probably not acceptable, since it confuses asan.  What I like about it is that it's almost branch free and should be pretty fast even for small inputs.  It doesn't work with the sanitizer build though :-(

This was done with help from Claude.  Transcript is at https://gist.github.com/erikcorry/e786ca8689fcacb30661deabb7258617

Type of change
- [ ] Bug fix
- [ x] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

### DIFF
--- a/src/icelake/icelake_base64.inl.cpp
+++ b/src/icelake/icelake_base64.inl.cpp
@@ -574,23 +574,21 @@ icelake_binary_length_from_base64(const char *input, size_t length) {
   const char *end = input + length;
 
   // Scalar prefix: process bytes until aligned to 64 bytes.
-  const char *aligned_ptr =
-      reinterpret_cast<const char *>((reinterpret_cast<size_t>(ptr) + 63) & ~63);
-  if (aligned_ptr > end) {
-    aligned_ptr = end;
-  }
-  while (ptr < aligned_ptr) {
+  size_t alignment_length = std::min(length, 64 - (reinterpret_cast<size_t>(ptr) & 63));
+  while (ptr < input + alignment_length) {
     count += (*ptr > 0x20) ? 1 : 0;
     ptr++;
+    length--;
   }
 
   // Process 64 bytes at a time with aligned loads.
   __m512i spaces = _mm512_set1_epi8(0x20);
-  while (ptr + 64 <= end) {
+  while (length >= 64) {
     __m512i data = _mm512_load_si512(reinterpret_cast<const __m512i *>(ptr));
     uint64_t mask = _mm512_cmpgt_epi8_mask(data, spaces);
     count += count_ones(mask);
     ptr += 64;
+    length -= 64;
   }
 
   // Scalar tail: process remaining bytes.
@@ -601,9 +599,8 @@ icelake_binary_length_from_base64(const char *input, size_t length) {
 
   // Count padding at the end.
   size_t padding = 0;
-  size_t pos = length;
-  while (pos > 0 && padding < 2) {
-    char c = input[--pos];
+  while (end > input && padding < 2) {
+    char c = *--end;
     if (c == '=') {
       padding++;
     } else if (c > ' ') {
@@ -621,23 +618,22 @@ icelake_binary_length_from_base64(const char16_t *input, size_t length) {
   const char16_t *end = input + length;
 
   // Scalar prefix: process char16_t until aligned to 64 bytes.
-  const char16_t *aligned_ptr = reinterpret_cast<const char16_t *>(
-      (reinterpret_cast<size_t>(ptr) + 63) & ~63);
-  if (aligned_ptr > end) {
-    aligned_ptr = end;
-  }
-  while (ptr < aligned_ptr) {
+  size_t alignment_length = std::min(length * 2, 64 - (reinterpret_cast<size_t>(ptr) & 63));
+  alignment_length /= 2;
+  while (ptr < input + alignment_length) {
     count += (*ptr > 0x20) ? 1 : 0;
     ptr++;
+    length--;
   }
 
   // Process 32 char16_t (64 bytes) at a time with aligned loads.
   __m512i spaces = _mm512_set1_epi16(0x20);
-  while (ptr + 32 <= end) {
+  while (length >= 32) {
     __m512i data = _mm512_load_si512(reinterpret_cast<const __m512i *>(ptr));
     __mmask32 mask = _mm512_cmpgt_epi16_mask(data, spaces);
     count += _mm_popcnt_u32(mask);
     ptr += 32;
+    length -= 32;
   }
 
   // Scalar tail: process remaining char16_t.
@@ -648,9 +644,8 @@ icelake_binary_length_from_base64(const char16_t *input, size_t length) {
 
   // Count padding at the end.
   size_t padding = 0;
-  size_t pos = length;
-  while (pos > 0 && padding < 2) {
-    char16_t c = input[--pos];
+  while (end > input && padding < 2) {
+    char16_t c = *--end;
     if (c == '=') {
       padding++;
     } else if (c > ' ') {


lemire:
@erikcorry I don't mind this but the reason I don't do it is that then we get bug reports due to sanitizers. So to make it usable, I think we need to have code to quiet the sanitizers... but I am totally open to discussion and I would not block your PR.

erikcorry:
Looks like I broke it on Windows with the custom version of popcount.

erikcorry:
> @erikcorry I don't mind this but the reason I don't do it is that then we get bug reports due to sanitizers. So to make it usable, I think we need to have code to quiet the sanitizers... but I am totally open to discussion and I would not block your PR.

We could tell the sanitizer to ignore this function, but perhaps it's a better fix to fall back to the scalar version when the sanitizer is enabled. This way the sanitizer can still find bugs in the rest of the program, for example passing an invalid length parameter to our function.

Of course if we do this we have to be 100% sure we don't have an actual bug in our implementation :-)

src/haswell/avx2_base64.cpp
```diff
@@ -768,3 +768,149 @@ class block64 {
     return 63;
   }
 };
+
+// The built-in version of _mm256_movemask_epi8 returns an int, but the
+// instruction actually returns a zero-extended 64 bit value.  The compiler
+// will do silly sign-extend and top-half-zeroing instructions because of this.
+// This version doesn't lie to the compiler about the result size.
```
lemire: I don't think we need these asm coded functions.

pauldreik:
what is the exact sanitizer error you get?

lemire:
@erikcorry 

What we do elsewhere in the simdutf code is to not worry about alignment, read the pointer, and handle the tail with scalar code (except with ISAs such as AVX-512 where that part can be done using scalar code. 

In some instances, we do use the alignment trick, but not to read out of bounds... that is, we still have a scalar tail.

https://github.com/simdutf/simdutf/blob/d56ea3e42314c2e9e403bde9f33ae4a7a456e234/src/generic/find.h#L6

Well, you do end up doing this scalar fixing at the end in any case.

I have a very long held view that optimizing for alignment is not usually important:

- Daniel Lemire, "Dot product on misaligned data," in Daniel Lemire's blog, July 14, 2025, https://lemire.me/blog/2025/07/14/dot-product-on-misaligned-data/.

For the 'find' function above (see first link), optimizing the alignment is likely beneficial... It might also be beneficial in the case of this function, given how simple it is... but I submit to you that we want to avoid the out-of-bound reads unless we can demonstrate tangible benefits.

Consider that even if we exclude the function from LLVM's sanitizers, for example, this will not free us from other sanitizers like valgrind and friends. There are others as well. I don't think we can ever be sure that we can detect the sanitizer. So there is some kind of cost here, and we need to make sure that it is really worth it to pay.

My instinct here is that the OOB behavior is unwarranted. (But I'd love to be proven wrong.)


As an aside, this implementation that you are doing can probably be done with our abstraction layer (see the first link in this comment).

erikcorry:
> Well, you do end up doing this scalar fixing at the end in any case.

Confused by this comment.  There's no scalar fixing in this change, it's all done with SIMD.

erikcorry:
> what is the exact sanitizer error you get?

For a test with a 5-byte buffer starting at 0xf00baa500 we read the 32 bytes from 0xf00baa500-0xf00baa1f.  So we are going over the end of the buffer.  But we don't count those bytes, and this can't cause a segfault because we are reading an aligned area that includes the buffer. Nobody has a 16 byte page size.

It's a legit sanitizer error, it's just harmless in the absence of sanitizers.

erikcorry:
> I have a very long held view that optimizing for alignment is not usually important

To be clear I don't think the alignment makes things faster for the bulk of a large operation, it just enables us to do the whole operation without falling back on a scalar loop at the end.

lemire:
@erikcorry 

You have a scalar patch up at `// Fix the end: subtract any over-counted bytes past 'end'`. It is relatively cheap, but still needed.

erikcorry:
OK wasn't counting that as 'scalar' because it subtracts all the bytes in one parallel operation rather than using a loop.

erikcorry:
Closing this because I don't have time to benchmark on small inputs, where it might be expected to help, and the disadvantages around sanitizers are clear.

For reference, here's the generated code.

https://godbolt.org/z/qKv4zx8bf

erikcorry:
I reopened and I'm trying to simplify.  For Haswell the simple version with the scalar cleanup is just as good.

Running on a 128k input file.

This is the Icelake speed on 0d55bddf no asm.
simdutf::icelake_binary_length_from_base64    :  124.41 GB/s  18.03 %   4.41 GHz   0.04 c/b   0.17 i/b   4.92 i/c 

With the latest e1a01e6e I only get:
simdutf::icelake_binary_length_from_base64    :  113.87 GB/s  15.96 %   4.17 GHz   0.04 c/b   0.13 i/b   3.52 i/c 

With the completely unaligned Icelake version 21bc1f1d I get miserable:
simdutf::icelake_binary_length_from_base64    :  80.85 GB/s  20.12 %   4.11 GHz   0.05 c/b   0.13 i/b   2.53 i/c 

I'm not sure why this is.

src/haswell/implementation.cpp
```diff
@@ -1406,6 +1406,18 @@ size_t implementation::binary_to_base64_with_lines(
   }
 }
 
+simdutf_warn_unused size_t
+implementation::binary_length_from_base64(const char *input,
+                                          size_t length) const noexcept {
```
anonrig, line 1411: since the methods are virtual, we need override here and below right?

src/haswell/avx2_base64.cpp
```diff
@@ -768,3 +768,80 @@ class block64 {
...
+  const char16_t *end = input + length;
+
+  // Process 16 char16_t (32 bytes) at a time.
+  // movemask gives 2 bits per char16_t, so we divide by 2 at the end.
+  __m256i spaces = _mm256_set1_epi16(0x20);
+  while (ptr + 16 <= end) {
+    __m256i data = _mm256_loadu_si256(reinterpret_cast<const __m256i *>(ptr));
+    __m256i gt_space = _mm256_cmpgt_epi16(data, spaces);
+    uint32_t mask = _mm256_movemask_epi8(gt_space);
+    count += count_ones(mask);
```
anonrig, line 824: mask is 32 bits set for 16 char16_t. This could potentially overflow won't it? Would it make sense to divide by 2 in here?

erikcorry: I don't see how it can overflow. Count is 64 bits.  We get twice as many ones as we need here, but we fix that below.

erikcorry:
The latest two versions are miscompiled for some reason - perhaps there's a bug in my code. Looking at this it's not surprising that it loses some speed.  Can't work out what is going on.

https://godbolt.org/z/qr88rf3x8

pauldreik:
> The latest two versions are miscompiled for some reason - perhaps there's a bug in my code. Looking at this it's not surprising that it loses some speed. Can't work out what is going on.
> 
> https://godbolt.org/z/qr88rf3x8

you have UB in the expressions `aligned_ptr += 64;` and `aligned_ptr < end` since it is illegal to step a pointer after the hypothetical element one past the end.

erikcorry:
The slowest version is the one with no aligned pointers so I don't think that's the issue.

src/icelake/icelake_base64.inl.cpp
```diff
@@ -564,3 +564,98 @@ compress_decode_base64(char *dst, const chartype *src, size_t srclen,
...
+    count += (*ptr > 0x20) ? 1 : 0;
+    ptr++;
+  }
+
+  // Process 64 bytes at a time with aligned loads.
+  __m512i spaces = _mm512_set1_epi8(0x20);
+  while (ptr + 64 <= end) {
+    __m512i data = _mm512_load_si512(reinterpret_cast<const __m512i *>(ptr));
+    uint64_t mask = _mm512_cmpgt_epi8_mask(data, spaces);
+    count += __builtin_popcountll(mask);
```
lemire:
```suggestion
    count += count_ones(mask);
```

lemire:
@erikcorry

I don't know the performance of your machine.  But if you try something like this...

```
./build/benchmarks/base64/benchmark_base64 -d somebase64file -f simdutf::icelake
```

You might get... I don't know... Let us say 20 GBs/s.

Ok. Suppose you compute the length at 80 GB/s... That's two passes for a net speed of 1/(1/80 +1/20) or 16 GB/s. 

That's reasonable, right?

No matter how we square this, doing two passes over the data is going to be slower. This cannot be helped. But if your penalty is, say, 25%... Then that's a reasonable trade-off.

src/icelake/icelake_base64.inl.cpp
lemire: (We define our own bit manip functions for portability issues. Got to love Microsoft.)

src/haswell/implementation.cpp
lemire: @anonrig It is a fair point, but I am not sure that we use the override keyword consistently in the code base.  Let me open an issue.

lemire:
@anonrig 
@erikcorry Please check https://github.com/simdutf/simdutf/pull/904

erikcorry:
> You might get... I don't know... Let us say 20 GBs/s.
> 
> Ok. Suppose you compute the length at 80 GB/s... That's two passes for a net speed of 1/(1/80 +1/20) or 16 GB/s.

The actual numbers for my Icelake (actually Sapphire Rapids) are 18GB/s and 130-140GB/s, so it's not too bad.  Something like a 13% penalty on a 128k input for having two passes, where the first one is very simple.

The latest version without UB (I think) is the fastest, but I think it's just a coincidence. Whenever I change something the compiler changes its strategy. I can't work out what the compiler is doing, but it's far from straightforward, and not faster than the always aligned sanitizer-hostile version was.

Without the scalar prologue section to get to an aligned point, the AVX512 version doesn't go nearly as fast (80GB/s only).

```
Icelake GB/s                  |  gcc-13.3 -O3  | clang-20 -O3
Version                       |                |
21bc1f1d No alignment         |   80.95        |  81.77
4d957af4 Aligned              |  108.90        | 109.49
6e83fc84 Aligned no UB        |  119.21        | 141.17
0d55bddf Aligned OOB reads    |  128.71        | 133.69
82604ca9 With inline ASM      |  123.84        | 130.65
```

lemire:
I can review later.

lemire:
We’ll get this done

erikcorry:
I had a quick stab at using the generic instead of the custom version for icelake and haswell, but I got a bit lost in the include changes needed, so I abandoned that, but I'd still be interested to see what the performance penalty is for using the generic implementation.

lemire:
@erikcorry I'm travelling but I will work later on this, to help finish it up.

The purpose is not replace the intrinsic code, but rather to scale our implementation over the many kernels.

Issue #911, opened 2026-01-12
# Heap buffer overflow in convert_utf16_to_utf8_safe


OwenSanzas:
## Summary
Heap-buffer-overflow at `utf16_to_utf8.h:141` due to "write-before-check" pattern in the 1-byte ASCII case. The `_safe` function designed to prevent buffer overflows actually causes one.

## Root Cause

**Two locations** with identical "write-before-check" bug:

### Bug 1: ASCII fast path (lines 122-131)
```cpp
while (pos < final_pos) {
  *utf8_output++ = ...;   // BUG: Write FIRST
  pos++;
  if (check_output && size_t(end - utf8_output) == 0) {  // Check AFTER
    return full_result(error_code::OUTPUT_BUFFER_TOO_SMALL, ...);
  }
}
```

### Bug 2: Single-char ASCII (lines 139-146)
```cpp
if ((word & 0xFF80) == 0) {
  *utf8_output++ = char(word);   // BUG: Write FIRST
  pos++;
  if (check_output && size_t(end - utf8_output) == 0) {  // Check AFTER
    return full_result(error_code::OUTPUT_BUFFER_TOO_SMALL, ...);
  }
}
```

### Correct pattern (2-byte case, lines 147-156):
```cpp
if (check_output && size_t(end - utf8_output) < 2) {  // Check FIRST
  return full_result(error_code::OUTPUT_BUFFER_TOO_SMALL, ...);
}
*utf8_output++ = ...;  // Write AFTER
```

Both 1-byte cases write before checking, causing overflow when buffer is exactly full.

### Minimal reproduction
```cpp
#include "simdutf.h"

int main() {
    // 'é' (2 UTF-8 bytes) + 'A' (1 UTF-8 byte) = 3 bytes needed
    char16_t input[] = {0x00E9, 'A'};
    char output[2];  // Only 2 bytes buffer

    // API says utf8_len is "the maximum output length"
    // Function should write at most 2 bytes, but writes 3 → overflow!
    size_t written = simdutf::convert_utf16_to_utf8_safe(
        input, 2, output, 2);

    // written == 3, violating the API contract
    return 0;
}
```

### Fuzzer harness (discovers this bug)

This is a custom fuzzer we wrote to test `_safe()` functions not covered by existing fuzzers:

```cpp
// fuzz_find_safe.cpp - Fuzzer for simdutf find() and *_safe() functions
#include <cstddef>
#include <cstdint>
#include <cstring>
#include <vector>
[...]

PR #912, opened 2026-01-12
Fix heap-buffer-overflow in convert_utf16_to_utf8_safe

OwenSanzas:
## Summary

- Move bounds check BEFORE write for 1-byte ASCII cases to prevent heap-buffer-overflow
- Fix two locations with the same "write-before-check" pattern (lines 123 and 141)
- The 2-byte, 3-byte, and 4-byte cases already correctly check before writing

## Root Cause

The `_safe` function designed to prevent buffer overflows had a bug where it writes to the buffer BEFORE checking if there is sufficient space:

```cpp
// BUG: write FIRST, check AFTER
*utf8_output++ = char(word);
pos++;
if (check_output && size_t(end - utf8_output) == 0) {  // Too late!
```

## Fix

Move the check before the write:

```cpp
// FIX: check FIRST, write AFTER
if (check_output && size_t(end - utf8_output) < 1) {
  return full_result(error_code::OUTPUT_BUFFER_TOO_SMALL, pos, utf8_output - start);
}
*utf8_output++ = char(word);
pos++;
```

## Test Plan

- [x] Verified fix with AddressSanitizer
- [x] Original crash input no longer triggers overflow
- [x] Normal conversion still works correctly

Fixes: #911

### DIFF
--- a/include/simdutf/scalar/utf16_to_utf8/utf16_to_utf8.h
+++ b/include/simdutf/scalar/utf16_to_utf8/utf16_to_utf8.h
@@ -120,14 +120,14 @@ simdutf_constexpr23 full_result convert_with_errors(InputPtr data, size_t len,
         if ((v & 0xFF80FF80FF80FF80) == 0) {
           size_t final_pos = pos + 4;
           while (pos < final_pos) {
+            if (check_output && size_t(end - utf8_output) < 1) {
+              return full_result(error_code::OUTPUT_BUFFER_TOO_SMALL, pos,
+                                 utf8_output - start);
+            }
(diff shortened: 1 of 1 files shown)


lemire:
Verification of the issue at https://github.com/simdutf/simdutf/pull/913

lemire:
@OwenSanzas Can you sync with our main branch to verify the fix ?

pauldreik:
I tried this PR on the newly written fuzzer just merged seconds ago. It works!

pauldreik:
The newly added tests for this also stopped failing with this patch applied. I suggest we merge it.

PR #913, opened 2026-01-12
verifies issue 911

lemire:
See  https://github.com/simdutf/simdutf/issues/911

### DIFF
--- a/tests/convert_utf16_to_utf8_safe_tests.cpp
+++ b/tests/convert_utf16_to_utf8_safe_tests.cpp
@@ -60,11 +60,9 @@ inline void verify_subset(std::vector<char16_t> &utf16,
 TEST(issue911) {
   char16_t input[] = {0x00E9, 'A'};
   char output[2];
-  size_t written = simdutf::convert_utf16_to_utf8_safe(
-      input, 2, output, 2);
-  ASSERT_EQUAL(written, 2);
+  size_t written = simdutf::convert_utf16_to_utf8_safe(input, 2, output, 2);
+  ASSERT_TRUE(written <= 2);
 }
-
 
 TEST(convert_pure_ASCII) {
   size_t counter = 0;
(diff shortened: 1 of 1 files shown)


tests/convert_utf16_to_utf8_safe_tests.cpp
```diff
@@ -57,6 +57,15 @@ inline void verify_subset(std::vector<char16_t> &utf16,
   }
 }
 
+TEST(issue911) {
+  char16_t input[] = {0x00E9, 'A'};
+  char output[2];
+  size_t written = simdutf::convert_utf16_to_utf8_safe(
+      input, 2, output, 2);
+  ASSERT_EQUAL(written, 2);
```
pauldreik: shouldn't this verify written<=2 ?

lemire: @pauldreik Hmmm... The first character in the input is `é` which should be converted to TWO bytes in UTF-8. No?

pauldreik:
I made a constexpr test for this, which also catches it:

```cpp
namespace {
template <auto input, std::size_t buflen>
constexpr auto convert_insufficient_buf() {
  using namespace simdutf::tests::helpers;
  constexpr auto Noutput = simdutf::utf8_length_from_utf16(input);
  CTString<char8_t, buflen> output{};
  const auto ret = simdutf::convert_utf16_to_utf8_safe(input, output);
  if (ret == 0) {
    throw "failed conversion";
  }
  return output;
}
} // namespace
TEST(compile_time_check_of_issue_911) {
  using namespace simdutf::tests::helpers;
  constexpr auto input = u"\u00E9A"_utf16;
  constexpr auto expected = u8"\u00E9A"_utf8;
  constexpr auto actual = convert_insufficient_buf<input, 2>();
  constexpr auto N = std::min(actual.size(), expected.size());
  static_assert(expected.shrink<N>() == actual.shrink<N>());
}
```

tests/convert_utf16_to_utf8_safe_tests.cpp
pauldreik: Yes, but I was thinking we are just verifying the promise of not writing past the output length.

lemire: I don't mind. Changing it.

lemire:
> I made a constexpr test for this, which also catches it:

Adding it.

lemire:
@pauldreik Have another look.

pauldreik (approved):

PR #915, opened 2026-01-12
fixing issue 914

lemire:
Fixes https://github.com/simdutf/simdutf/issues/914


lemire:
I'll merge because it is low-risk.

PR #917, opened 2026-01-12
Release candidate 8.0.0

lemire:


pauldreik:
great that you make a new release so fast after the write buffer overflow was found and fixed!

lemire:
> great that you make a new release so fast after the write buffer overflow was found and fixed!

Yeah. We have a bunch of new stuff to release. The only thing I am unsure about is whether it should be a major release.

pauldreik:
> > great that you make a new release so fast after the write buffer overflow was found and fixed!
> 
> Yeah. We have a bunch of new stuff to release. The only thing I am unsure about is whether it should be a major release.

Tooting my own horn here, but I think the constexpr thing is a pretty major thing.

lemire:
@pauldreik Yeah. I think we need to make it 8.0.0.

But the tests need to go green first.

Issue #921, opened 2026-01-15
Unit test failure when building the release tarball: check_feature_macros


clausecker:
The unit test *check_feature_macros* of simdutf 8.0.0 fails on aarch64 and armv7 FreeBSD 15.0 built with the system compiler, clang 19.1.7.  This'll probably reproduce on other configurations as well.  I am building from the release tarball, so no git repository.  The unit test fails as follows (from `LastTest.log`):

```
89/89 Testing: check_feature_macros
89/89 Test: check_feature_macros
Command: "/usr/local/bin/python3.11" "/usr/home/ports/main.ports/converters/simdutf/work/simdutf-8.0.0/scripts/check_feature_macros.py"
Directory: /usr/home/ports/main.ports/converters/simdutf/work/simdutf-8.0.0
"check_feature_macros" start time: Jan 15 17:30 CET
Output:
----------------------------------------------------------
Error running git ls-files: Command '['git', 'ls-files', '*.cpp', '*.h']' returned non-zero exit status 128.
No .cpp or .h files found in the Git repository.
<end of output>
Test time =   0.20 sec
----------------------------------------------------------
Test Failed.
"check_feature_macros" end time: Jan 15 17:30 CET
"check_feature_macros" time elapsed: 00:00:00
----------------------------------------------------------
```

It seems like the unit test should either be skipped if no git repository is used or be modified to no longer require git.

lemire:
Yeah. Good catch. Let us not fail so easily.

We still want this test around to nag us because the feature macros can easily get mixed up when people try to contribute, but it should not be causing false alarms.

pauldreik:
can we verify this in ci? I get a bit of deja vu, didn't we already go through this?

lemire:
@pauldreik A test in CI would be nice. Issue: https://github.com/simdutf/simdutf/issues/923

PR #922, opened 2026-01-15
fixing issue 921

lemire:
When git fails to return the files (i.e., when we are not under git control), we should just have the `check_feature_macros` test succeed (with a warning message).

Fixes: https://github.com/simdutf/simdutf/issues/921

### DIFF
--- a/scripts/check_feature_macros.py
+++ b/scripts/check_feature_macros.py
@@ -131,7 +131,7 @@ def main():
     files = get_git_files(args.limit)
     if not files:
         print("No .cpp or .h files found in the Git repository" + (" in src/ or include/" if args.limit else "") + ".")
-        sys.exit(0)  # Exit with success code if no files are found, see https://github.com/simdutf/simdutf/issues/921
+        sys.exit(142)  # Exit with special code if no files are found, see https://github.com/simdutf/simdutf/issues/921
 
     all_errors = []
     statistics = {'if': 0, 'ifdef': 0, 'ifndef': 0, 'endif': 0}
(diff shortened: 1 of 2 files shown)


clausecker:
Does the test framework provide a mechanism to skip unit tests?  Perhaps that would be more appropriate.

lemire:
> Does the test framework provide a mechanism to skip unit tests?

I am not sure what you mean. You definitely can skip running the tests. You can also skip a specific unit test `ctest -E "^exactly_my_test_name$"`.

If you mean skipping this one test when the files are not under git... it is essentially what my PR does, it tries to recover the files, but if it can't, it just does not fail the test.

clausecker:
Some unit test frameworks have a SKIP result for unit tests in addition to the usual PASS/FAIL ones.  SKIP means “test was not run and we cannot say if it would pass or not.”  This is more meaningful than just having the unit test pass, as it makes clear that nothing was actually tested.

CMake-based tests support this through the [SKIP_RETURN_CODE](https://cmake.org/cmake/help/latest/prop_test/SKIP_RETURN_CODE.html) variable.  Maybe that is applicable here?

scripts/check_feature_macros.py
```diff
@@ -131,7 +131,7 @@ def main():
     files = get_git_files(args.limit)
     if not files:
         print("No .cpp or .h files found in the Git repository" + (" in src/ or include/" if args.limit else "") + ".")
-        sys.exit(1)  # Exit with error code if no files are found
+        sys.exit(0)  # Exit with success code if no files are found, see https://github.com/simdutf/simdutf/issues/921
```
clausecker:
```suggestion
        sys.exit(2)  # Exit with success code if no files are found, see https://github.com/simdutf/simdutf/issues/921
```

And then in tests/CMakeLists.txt, set the SKIP_RETURN_CODE property of this test to 2.

lemire:
@clausecker Yes, well ahead of you. Check my last commit.

I used 142 instead of 2 because 142 is 2 * 71 which is obviously good.

clausecker:
I like it!  LGTM.

lemire:
@clausecker I will let the tests go green.

PR #931, opened 2026-02-05
This adds 'override' annotations for RVV, lasx and lsx.

lemire:
The override annotation is not super useful in my experience, but it does add a bit of clarity to the code.


pauldreik:
looks good! pushed a commit with the formatting applied.

pauldreik:
on gitlab, there is a button "merge when CI is green". that button would be great on github too.

lemire:
I'll merge eagerly. I don't expect that you broke the build.

PR #932, opened 2026-02-05
inlining get_default_implementation and get_single_implementation

lemire:
The the get_default_implementation and get_single_implementation functions should be inlined, for best performance. (They appear in the profiling data.)


lemire:
> did you measure any performance difference?

Yes, maybe 5% of 10% difference on very short calls in some instances.

Issue #925, opened 2026-01-22
add short inputs optimizations


lemire:
When processing short inputs (e.g., strings smaller than 16 bytes), calling our fast functions is wasted effort. Now that @pauldreik has done the hard work of moving our scalar (=naive) implementations in header files, what becomes possible is to add short string optimizations: when the input is sufficiently small, prefer the naive (but easy to inline) routine.

It can be an actual issue. Recently @anonrig tried to enable simdutf in ada, within Node, and he reported a negative effect on performance. Though I have not examined the use case, the likely cause in my view is that he is replacing a trivial (small, pure C function) with a non-trivial function call into simdutf.

You can concrete examples in Node...

https://github.com/nodejs/node/blob/e9b0849606ccaa607698ccb2e88f7245113968c3/src/node_buffer.cc#L760-L764 

https://github.com/nodejs/node/blob/e9b0849606ccaa607698ccb2e88f7245113968c3/src/string_bytes.cc#L542-L559

(The latter example is from @ChALkeR whereas I think that the first one is from me.)

My expectation is not that the actual implementation in simdutf are inefficient for short strings, but rather that replacing a less efficient, but inlinable, function, with a non-inlineable function that is faster (as long as it has enough work to do) can be a slight net negative on short inputs.

With the current code base, it *should* now be relatively easy to add, directly in the simdutf header file, something like this ...

```
if(the input is short) {
  // call the simple scalar function
} else {
  // dispatch into the simdutf lib for the optimized function
}
```

Obviously, this entails adding relevant benchmarks, which we do not have right now.

ronaksarda:
Hi, I’d like to work on this issue.

My understanding is that for very short inputs, the overhead of dispatching into the SIMD implementation can outweigh the benefits, especially compared to an inlinable scalar implementation.

I’m planning to add a small length-based dispatch in the public validate_utf8 entry point so that short inputs use the existing scalar path, starting with a conservative threshold (e.g., 32 bytes), and leaving room for tuning via benchmarks later.

Please let me know if this approach sounds reasonable.

lemire:
Yes. I do invite contributions!

For the benchmarks, we already have them in a sense: the `benchmark` utility would do, we just need to pass a file that is short.

> My understanding is that for very short inputs, the overhead of dispatching into the SIMD implementation can outweigh the benefits, especially compared to an inlinable scalar implementation.

[...]

PR #935, opened 2026-02-07
Inline short-input fast path for validate_utf8 with dispatch fallback

ronaksarda:
Add an inline fast path for validate_utf8(const char*, size_t)
in the public header, following the direction discussed in #925.

For very small inputs, the wrapper directly calls the scalar UTF-8
validator so the short-input path can be inlined and avoid the fixed
overhead of the non-inline entry point.

For larger inputs, behavior is unchanged: the call is forwarded to a
non-inline dispatch helper implemented in src/implementation.cpp,
preserving existing runtime dispatch and architecture-specific
optimizations.

Local shortbench measurements on an Ice Lake i7-1165G7 show roughly
2× lower latency for 1-byte inputs (~5.5 ns → ~2.4 ns), with no
observable regressions for larger sizes.

All existing tests pass locally. No behavioral changes are intended.

### DIFF
--- a/benchmarks/shortbench.cpp
+++ b/benchmarks/shortbench.cpp
@@ -1,19 +1,14 @@
-// This benchmnark program seeks to measure the performance of very short
-// functions calls to various simdutf functions.
-// You can process the result with the script at scripts/shortinputplots.py
-// to generate plots. See the instructions in that script for more details.
 #include <algorithm>
-#include <array>
-#include <atomic>
-#include <cmath>
 #include <cstddef>
-#include <cstring>
 #include <fstream>
-#include <functional>
 #include <iostream>
-#include <span>
-#include <stdexcept>
 #include <vector>
+#include <stdexcept>
+#include <atomic>
+#include <array>
+#include <cstring>
+#include <cmath>
+#include <functional>
 
 #include "simdutf.h"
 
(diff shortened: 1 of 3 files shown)


ronaksarda:
Hi @lemire,

Just a gentle follow-up on this PR.

This change adds an inline fast path for very small inputs to
validate_utf8, based on the direction discussed in #925, while
preserving the existing dispatch behavior for larger inputs.

All tests pass locally, and I’m happy to adjust the implementation
or run additional benchmarks if needed.

Thanks for your time.

lemire:
@ronaksarda Thanks for the PR.

At this time, we are still researching the issue. I do not think we will make changes to simdutf before we have an in-depth understanding of our options and a careful validation. We don't want to introduce a breaking change without careful consideration.

ronaksarda:
Thanks for the clarification and for looking into this.

That makes sense, especially given the potential ABI and runtime
implications. I agree that careful validation and a solid understanding
of the available design options are important before introducing such
a change.

I’m happy to help with additional experiments, benchmarks, or
prototyping if that would be useful during the investigation.

Thanks again for the guidance.

ronaksarda:
Thanks for the detailed feedback and for taking the time to review this.

I also noticed that the current CI matrix reports widespread failures,
which suggests the inline change at the public header boundary may have
broader build or ABI implications than initially expected.

Given this, I agree that further investigation and design discussion are
likely needed before pursuing this direction.

I’m happy to help with additional experiments or alternative approaches
if that would be useful, and I appreciate the guidance.

PR #936, opened 2026-02-07
add convert_utf16_to_utf8_with_replacement

mertcanaltin:
utf8_length_from_utf16_with_replacement was already added in v7.7.0, but the actual conversion function was missing. This PR adds it — converting directly to UTF-8 in a single call, without an intermediate buffer, by replacing broken surrogates with U+FFFD.

In Node.js, in encoding_binding.cc, in places like TextEncoder.encodeInto and similar locations, the string received from the user arrives in UTF-16 format, but it may contain corrupted surrogates You need to convert this to UTF-8, but convert_utf16_to_utf8 throws an error when it encounters a corrupted surrogate.

[  - node/src/encoding_binding.cc line 144](https://github.com/mertcanaltin/node/blob/main/src/encoding_binding.cc#L144)
[  - node/src/encoding_binding.cc line 283](https://github.com/mertcanaltin/node/blob/main/src/encoding_binding.cc#L283)

@anonrig @lemire

### DIFF
--- a/tests/convert_utf16_to_utf8_with_replacement_tests.cpp
+++ b/tests/convert_utf16_to_utf8_with_replacement_tests.cpp
@@ -247,7 +247,7 @@ template <auto input> constexpr auto invoke_convert_with_replacement() {
   } else {
     written = simdutf::convert_utf16be_to_utf8_with_replacement(input, output);
   }
-  if (written != 5) {
+  if (written != output.size()) {
     throw "unexpected length";
   }
   return output;


mertcanaltin:
I will inspect ubuntu s390x problem

lemire:
I think that's a great PR. We'll just have to continue the work with later PRs by providng optimized functions (per kernel). We should also add it to the fuzzer. This can be done later.

mertcanaltin:
The s390x failure was a big-endian issue in the tests. The hardcoded char16_t values (like {0xD83D, 0xDE00}) are stored in native byte order, but convert_utf16le_to_utf8_with_replacement expects little-endian data in memory. On s390x the bytes are swapped, so the function sees regular BMP characters instead of a surrogate pair. Fixed by wrapping test inputs with to_utf16le(), same pattern used by the other UTF-16LE tests.

tests/convert_utf16_to_utf8_with_replacement_tests.cpp
```diff
@@ -0,0 +1,211 @@
...
+  ASSERT_TRUE(std::memcmp(output.data(), fffd_utf8, 3) == 0);
+  ASSERT_EQUAL(output[3], 'A');
+}
+
+// Test: unpaired low surrogate should be replaced with U+FFFD
+TEST(unpaired_low_surrogate_le) {
+  // DC00 (low surrogate without preceding high surrogate)
+  std::vector<char16_t> input = {to_utf16le(u'B'),
+                                 to_utf16le(char16_t(0xDC00)),
+                                 to_utf16le(u'C')};
```
mertcanaltin:
```suggestion
  std::vector<char16_t> input = {to_utf16le(u'B'), to_utf16le(char16_t(0xDC00)),
                                 to_utf16le(u'C')};
```

pauldreik (changes requested):
I would like to see at least one constexpr test, so we know it is callable in a constexpr context.

mertcanaltin:
> I would like to see at least one constexpr test, so we know it is callable in a constexpr context.

Sure, I will send today

lemire:
running tests.

mertcanaltin:
> running tests.

Sorry,  I fixed a minor linter error. I think We'll need to restart the tests.

lemire:
Tests restarted.

mertcanaltin:
They all succeeded.

lemire:
Let us give @pauldreik some time to get back to this issue.

tests/convert_utf16_to_utf8_with_replacement_tests.cpp
```diff
@@ -207,4 +211,66 @@ TEST_LOOP(output_is_valid_utf8_le) {
...
+template <auto input> constexpr auto invoke_convert_with_replacement() {
+  // 'A' = 1 byte, unpaired surrogate = 3 bytes (U+FFFD), 'B' = 1 byte
+  simdutf::tests::helpers::CTString<char, 5> output{};
+  std::size_t written;
+  if constexpr (decltype(input)::endianness == std::endian::little) {
+    written = simdutf::convert_utf16le_to_utf8_with_replacement(input, output);
+  } else {
+    written = simdutf::convert_utf16be_to_utf8_with_replacement(input, output);
+  }
+  if (written != 5) {
```
pauldreik: suggest: output.size() instead of 5.

src/ppc64/implementation.cpp
```diff
@@ -777,6 +777,20 @@ implementation::utf8_length_from_utf16be_with_replacement(
       endianness::BIG>(input, length);
 }
 
+simdutf_warn_unused size_t
+implementation::convert_utf16le_to_utf8_with_replacement(
+    const char16_t *input, size_t length, char *utf8_buffer) const noexcept {
```
pauldreik: missing override?

src/lsx/implementation.cpp
```diff
@@ -1201,6 +1201,20 @@ implementation::utf8_length_from_utf16be_with_replacement(
       endianness::BIG>(input, length);
 }
 
+simdutf_warn_unused size_t
+implementation::convert_utf16le_to_utf8_with_replacement(
+    const char16_t *input, size_t length, char *utf8_buffer) const noexcept {
```
pauldreik: missing override?

src/lasx/implementation.cpp
```diff
@@ -1315,6 +1315,20 @@ implementation::utf8_length_from_utf16be_with_replacement(
       endianness::BIG>(input, length);
 }
 
+simdutf_warn_unused size_t
+implementation::convert_utf16le_to_utf8_with_replacement(
+    const char16_t *input, size_t length, char *utf8_buffer) const noexcept {
```
pauldreik: missing override?

src/icelake/implementation.cpp
```diff
@@ -1680,6 +1680,20 @@ implementation::utf8_length_from_utf16be_with_replacement(
       input, length);
 }
 
+simdutf_warn_unused size_t
+implementation::convert_utf16le_to_utf8_with_replacement(
+    const char16_t *input, size_t length, char *utf8_buffer) const noexcept {
```
pauldreik: missing override?

pauldreik (approved):
this looks good! just a minor thing, but fine to merge anyway.

tests/convert_utf16_to_utf8_with_replacement_tests.cpp
mertcanaltin: thanks

mertcanaltin: https://github.com/simdutf/simdutf/pull/936/commits/72e77a3cf7787b922465e9f5a446bd05dbb62c70

mertcanaltin:
Solved, thanks for review

PR #937, opened 2026-02-08
Add `simdutf_utf8_length_from_utf32` to C API

triallax:
Short title (summary): Add `simdutf_utf8_length_from_utf32` to C API

Description
- What did you change and why? (1-3 sentences) 
  Added `simdutf_utf8_length_from_utf32` to the C API, it appears to have been missed when the header was introduced
- Issue reproduced / related issue: None

Type of change
- [ ] Bug fix
- [ ] Optimization
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

Checklist before submitting
- [x] I added/updated tests covering my change (if applicable)
  Don't think adding a test is necessary as this is a simple wrapper
- [x] Documentation / README updated if needed
  Don't think this is needed
- [x] I linked the related issue (if applicable)
       None

### DIFF
--- a/src/simdutf_c.cpp
+++ b/src/simdutf_c.cpp
@@ -145,6 +145,10 @@ size_t simdutf_utf8_length_from_utf16(const char16_t *input, size_t length) {
   return simdutf::utf8_length_from_utf16(
       reinterpret_cast<const char16_t *>(input), length);
 }
+size_t simdutf_utf8_length_from_utf32(const char32_t *input, size_t length) {
+  return simdutf::utf8_length_from_utf32(
+      reinterpret_cast<const char32_t *>(input), length);
+}
 simdutf_result
 simdutf_utf8_length_from_utf16_with_replacement(const char16_t *input,
                                                 size_t length) {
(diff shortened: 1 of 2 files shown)


src/simdutf_c.cpp
```diff
@@ -145,6 +145,10 @@ size_t simdutf_utf8_length_from_utf16(const char16_t *input, size_t length) {
   return simdutf::utf8_length_from_utf16(
       reinterpret_cast<const char16_t *>(input), length);
 }
+size_t simdutf_utf8_length_from_utf32(const char32_t *input, size_t length) {
+  return simdutf::utf8_length_from_utf32(
+      reinterpret_cast<const char32_t *>(input), length);
```
triallax, line 150: I added this cast here because it's used in some other functions like `simdutf_utf8_length_from_utf16_with_replacement`, but I wonder if it's really necessary? In both this and the other functions the cast is to the same type of `input` so it seems unnecessary to me

pauldreik: I think the original code was made with AI.

I agree this is unnecessary, please remove it (and at the other places too, if you have time).

PR #939, opened 2026-02-12
add AI policy

lemire:
Self-explanatory.

### DIFF
--- a/AI_USAGE_POLICY.md
+++ b/AI_USAGE_POLICY.md
@@ -32,6 +32,12 @@ unreviewed output of an LLM to open source project maintainers *extracts* work
 from them in the form of design and code review, so we call this kind of
 contribution an "extractive contribution".
 
+## Transparency
+
+For contributions involving significant AI assistance, we encourage you to disclose 
+its use and explain your process. If a submission appears to rely heavily on AI 
+without disclosure, we may doubt that the **human-in-the-loop** requirement has
+been met. Please show awareness of your use of AI.
 
 ## Copyright
 


pauldreik (approved):
well written and I agree!

I think it also is a good idea that PRs made partially or entirely by AI should mention that in the PR.  perhaps add it to the PR template?

lemire:
> PRs made partially or entirely by AI should mention that in the PR.

Do you have a wording in mind ?

AI_USAGE_POLICY.md
```diff
@@ -0,0 +1,50 @@
...
+## Extractive Contributions
+
+The reason for our "human-in-the-loop" contribution policy is that processing
+patches, PRs, RFCs, and comments is not free -- it takes a lot of
+maintainer time and energy to review those contributions! Sending the
+unreviewed output of an LLM to open source project maintainers *extracts* work
+from them in the form of design and code review, so we call this kind of
+contribution an "extractive contribution".
+
+
```
pauldreik, line 41: suggested addition:

## Transparency

When a PR is made partially or completely using AI, it is encouraged to indicate this in the PR.

lemire: @pauldreik Ok. So I added something. Please have a look.

I don't want people to disclose all uses of AI as that would be just silly (i.e., AI can include a vaste range of tools from checkers, autocomplete, and so forth). What we really care about, I think, are extensive use of AI [I prompted Claude and got this back]. We still want to accept these 'extensive use of AIs' but I think that they should be disclosed so that we know what to expect.

In other projects, I am getting wild bug reports and patches that are obviously entirely AI generated without a human in the loop. These people are not doing themselves a favour as I can quickly tell what is happening.

At the other extreme, if you typed you code, and an AI autocorrected your typo, I don't want to know about it.

So there is a spectrum and, in my view, we only care about the extreme end of the spectrum where a lot of work was done by the AI. These cases are still fine, but it is helpful of he author is upfront about it.

pauldreik:
the text looks great, let's merge!

Issue #908, opened 2026-01-10
make fastbase64 functionally equivalent to the standard base64 command


lemire:
Currently the fastbase64 command line utility looks like the standard base64 utility but it is missing some functionality like the ability to output line separated code (we have the corresponding function in the library, it is just needs to be used). The code could also be improved as the utility was written before the recent updates to the base64 functions.

pauldreik:
yes! when this is done we could perhaps come up with instructions how to replace the system base64 with fastbase64 instead. I just measured the performance, it is a whopping **7x faster** than coreutils in encode and **10x faster** in decode.

on debian based systems, one can use the /etc/alternatives system but I suspect that requires cooperation from the coreutils package.

lemire:
> I just measured the performance, it is a whopping 7x faster than coreutils in encode and 10x faster in decode.

I have issued PR
https://github.com/simdutf/simdutf/pull/947

I'd be interested in the benchmark that delivered 7x and 10x perf boosts. How large was the file?

pauldreik:
this is how I measured:
```plain
$ /usr/bin/time build/release/tools/fastbase64  --encode <10Ge >/dev/null
0.60user 0.96system 0:01.57elapsed 99%CPU (0avgtext+0avgdata 2520maxresident)k
0inputs+0outputs (0major+139minor)pagefaults 0swaps
$ /usr/bin/time base64  <10Ge >/dev/null
8.89user 1.74system 0:10.65elapsed 99%CPU (0avgtext+0avgdata 2068maxresident)k
0inputs+0outputs (0major+128minor)pagefaults 0swaps
$ ls -lah 10Ge 
-rw-rw-r-- 1 pauldreik pauldreik 13G 11 jan 21.21 10Ge
```

lemire:
@pauldreik Ok. Let me try this.

Issue #907, opened 2026-01-10
provide man pages for fastbase64 and sutf


pauldreik:
The man pages do not need to be perfect, just something to get a quick understanding of what the tool is and how to use it. They can be polished over time.

For reference, see the debian policy: https://www.debian.org/doc/debian-policy/ch-docs.html#manual-pages  (not that this project is bound by the debian policy, but it is still a good idea to have a man page)

lemire:
I agree.

PR #947, opened 2026-03-13
Better fastbase64 + man pages

lemire:
This makes fastbase64 somewhat equivalent to standard base64. Unfortunately, the GNU base64 and the non-GNU base64 tools differ. It tried to come up with some kind of hybrid that can emulate the two.

Apple M4 Benchmark Results (times in milliseconds):

Size     | Encode Base64 | Encode FastBase64 | Decode Base64 | Decode FastBase64
---------|---------------|-------------------|---------------|------------------
1m       | 23.5          | 23.2              | 37.7          | 23.1            
10m      | 31.3          | 24.4              | 167.6         | 24.0            
100m     | 115.6         | 50.4              | 1430.2        | 53.7  

x64 Linux Benchmark Results (times in milliseconds):

Size     | Encode Base64 | Encode FastBase64 | Decode Base64 | Decode FastBase64
---------|---------------|-------------------|---------------|------------------
1m       | 12.1          | 11.8              | 13.8          | 11.8            
10m      | 25.8          | 18.7              | 39.6          | 17.5            
100m     | 187.0         | 100.5             | 290.3         | 77.7            

So for inputs that are 10 MB or larger, it pays to use fastbase64. For small inputs, the time is dominated by the process overhead. 

The Apple base64 decoder is quite slow, so that's one use case where we may want to use fastbase64.

Fixes https://github.com/simdutf/simdutf/issues/907

Fixes https://github.com/simdutf/simdutf/issues/908


tools/fastbase64.cpp
```diff
@@ -19,71 +24,131 @@ class CommandLine {
...
+  bool write_with_wrapping(std::FILE *fp, const char *data, size_t length,
+                           int &current_col, int wrap_cols);
   static void show_help();
   bool decode;
 };
 
 CommandLine CommandLine::parse_and_validate_arguments(int argc, char *argv[]) {
   CommandLine cmdline;
-  std::vector<std::string> arguments;
+  cmdline.decode = true; // decode is default
```
pauldreik: suggest writing this as a member initializer at row 30 instea<d.

tools/fastbase64.cpp
```diff
@@ -19,71 +24,131 @@ class CommandLine {
...
+  cmdline.input_file = "-";
+  cmdline.output_file = "-";
+  cmdline.wrap_cols = 0;
+  cmdline.ignore_garbage = false;
```
pauldreik: these are already set as member initializers in row 11-16 which is good!

suggest removing these lines entirely.

tools/fastbase64.cpp
```diff
@@ -8,7 +8,12 @@
 class CommandLine {
 public:
   std::FILE *current_file{NULL};
-  std::filesystem::path output_file;
```
pauldreik, line 11: now with filesystem not being used anymore, please remove the include.

tools/fastbase64.cpp
```diff
@@ -223,18 +326,35 @@ bool CommandLine::encode_to(std::FILE *fpout) {
...
   return true;
 }
 
 void CommandLine::show_help() {
   printf("Usage: fastbase64 [OPTIONS...] [INPUTFILE] [OUTPUTFILE]\n\n");
-  printf("  -d       decode base64\n"
-         "  -e       encode base64 (default)\n\n");
-  printf("If output is not specified, the output is redirected to standard "
+  printf(
+      "  -b, --break NUM   break encoded output up into lines of length NUM\n");
```
pauldreik: suggest appending: (default 0, meaning no breaks)

(also applies to the man page)

tools/fastbase64.cpp
```diff
@@ -19,71 +24,131 @@ class CommandLine {
...
+      i += 2;
+    } else if (arg == "-o" || arg == "--output") {
+      if (i + 1 >= argc) {
+        throw std::runtime_error("Missing value for " + arg);
       }
+      cmdline.output_file = argv[i + 1];
+      i += 2;
+    } else if (arg == "--version") {
+      printf("fastbase64 version %s\n", SIMDUTF_VERSION);
+      exit(0);
```
pauldreik: std::exit(EXIT_SUCCESS)

tools/fastbase64.cpp
```diff
@@ -19,71 +24,131 @@ class CommandLine {
...
-        cmdline.output_file = a;
-      } else {
-        throw std::runtime_error("Too many arguments!");
+      i++;
+    } else if (arg == "--ignore-garbage") {
+      cmdline.ignore_garbage = true;
+      i++;
+    } else if (arg == "-h" || arg == "--help") {
+      show_help();
+      exit(0);
```
pauldreik: std::exit(EXIT_SUCCESS)

tools/fastbase64.cpp
pauldreik: these are already set in the member initializers in the class declaration, these lines should be removed.

tools/fastbase64.cpp
```diff
@@ -19,71 +24,131 @@ class CommandLine {
   std::pair<bool, size_t> load_chunk(char *input_data, size_t chunk_size,
                                      size_t offset);
   bool write_to_file_descriptor(std::FILE *fp, const char *data, size_t length);
+  bool write_with_wrapping(std::FILE *fp, const char *data, size_t length,
+                           int &current_col, int wrap_cols);
   static void show_help();
   bool decode;
```
pauldreik: =true; // decode is default

scripts/base64bench.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
pauldreik, line 1: did you consider writing this in python instead?

pauldreik:
I think it would be great to provide a drop in replacement for gnu coreutils.
We could provide a bash script wrapper `fastbase64.coreutilscompat` that parses command line parameters and uses defaults that exactly matches gnu coreutils.

then, it would be possible for users to either use an alias in .bashrc to get some nice speedup.

the same thing could be done for mac/bsd, I assume?

tools/fastbase64.cpp
lemire: done.

tools/fastbase64.cpp
lemire: good catch

tools/fastbase64.cpp
lemire: good.

tools/fastbase64.cpp
lemire: Fixed.

tools/fastbase64.cpp
lemire: Fixed.

tools/fastbase64.cpp
lemire: Done.

tools/fastbase64.cpp
lemire: Agreed.

scripts/base64bench.sh
lemire: We could, but I don't expect most people to use this script. :-)

lemire:
@pauldreik 

> I think it would be great to provide a drop in replacement for gnu coreutils.

I *think* that the only difference is the meaning of the flag `-i`.

We have a few options.

1. Add an intermediate script as you suggest.
2. Generate two binaries, one with GNU commands, one with BSD convention.
3. Focus solely on the GNU convention.
4. Stick with what I did which is mostly cross-compatible except that Linux users must not use `-i`.
5. Detect a GNU (Linux) system and change the behaviour under Linux so that `-i` means ignore bad characters.

lemire:
I am asking for advice on X : https://x.com/lemire/status/2035726774919713271?s=61&t=v2gDAuOzz1C3ICrzUYCSlQ

lemire:
This answer is interesting: https://x.com/austinsnotweird/status/2035730879461658876?s=61&t=v2gDAuOzz1C3ICrzUYCSlQ

chadbrewbaker:
I would defer to @sylvestre. The last Rust coreutils PR was for arg[0] handling: https://github.com/uutils/coreutils/commit/6c3de3e92fff02ef7212c7bb8730fc5357ee3da1

--bsdmode, --gnumode seem like the most sensible flags for a utility that isn't GNU  or BSD?

pauldreik:
I think busybox is also using the trick of checking argv[0] to let the same binary masquerade as different programs.

In the case of providing a symlink with a name of say fastbase64.coreutils which links to fastbase64, argv[0] is "/path/to/fastbase64.coreutils" (depends on how the link is invoked, one needs to do basename on it)

I think some people use gnu coreutils on mac, so it makes sense to provide the .coreutils link not only on linux.

I made a quick test of trying to make a symlink with cmake, using:

```cmake
install(CODE "execute_process( \
    COMMAND echo ${CMAKE_COMMAND} -E create_symlink \
    ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/fastbase64 \
    ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/fastbase64.coreutils   \
    )"
)
```

it does not work properly, I think a generator expression would have been ideal but is does not seem like it is supported: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#export-and-install-expressions 

there is a potentially useful hint here but I failed understanding how: https://discourse.cmake.org/t/how-to-properly-install-a-symbolic-link-for-backwards-compatibility/8592

lemire:
@pauldreik Please see

https://github.com/simdutf/simdutf/pull/949

I just call ln directly *or* copy the file as a fallback when ln is not available.

Now, I wonder about the name. Is `fastbase64.coreutils` what we want to go with?

lemire:
@pauldreik Ok, now the PR does have two utilities.

tools/fastbase64.cpp
```diff
@@ -2,92 +2,235 @@
...
+        i++;
       } else {
-        throw std::runtime_error("Too many arguments!");
+        if (i + 1 >= argc) {
+          throw std::runtime_error("Missing value for " + arg);
+        }
+        cmdline.input_file = argv[i + 1];
+        i += 2;
+      }
+    } else if (arg == "--input") {
```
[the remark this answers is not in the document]
lemire: I think that's ok.

lemire:
@pauldreik This should be in pretty decent shape now.

man/fastbase64.coreutils.1
```diff
@@ -0,0 +1,56 @@
...
+By default, it encodes binary input to base64 output. On large files,
+it can be several times faster than standard base64 tools.
+.PP
+.B fastbase64 extension (not present in GNU coreutils base64):
+If a second filename is provided, it is used as the output file.
+Otherwise, output goes to standard output. The \fB\-o\fR option takes precedence over positional output file.
+.SH OPTIONS
+.TP
+.B \-b, \-\-break=\fINUM\fR
+fastbase64 extension (not present in GNU coreutils base64): break encoded output into lines of length \fINUM\fR (default 76, meaning breaks at 76 characters).
```
pauldreik: suggest:
"synonym to --wrap. (not in GNU coreutils base64)"

man/fastbase64.coreutils.1
```diff
@@ -0,0 +1,56 @@
...
+Decode input.
+.TP
+.B \-e, \-\-encode
+Encode input (default).
+.TP
+.B \-\-ignore\-garbage
+When decoding, ignore non-alphabet characters.
+.TP
+.B \-i, \-n, \-\-noerrcheck
+When decoding, ignore non-alphabet characters (same as \-\-ignore\-garbage).
```
pauldreik: suggest appending: (not in GNU coreutils base64)

man/fastbase64.coreutils.1
```diff
@@ -0,0 +1,56 @@
...
+.TP
+.B \-h, \-\-help
+Display help message and exit.
+.TP
+.B \-o, \-\-output=\fIFILE\fR
+Output file (default: "-" for stdout).
```
pauldreik: suggest appending: (not in GNU coreutils base64)

tools/fastbase64.cpp
```diff
@@ -2,92 +2,251 @@
...
 #include <array>
 #include <cerrno>
+#include <cstring>
 #include <filesystem>
+#include <memory>
 #include <vector>
 
+class ArgumentError : public std::runtime_error {
+public:
+  ArgumentError(const std::string &msg) : std::runtime_error(msg) {}
```
pauldreik: explicit

tools/fastbase64.cpp
```diff
@@ -223,28 +448,70 @@ bool CommandLine::encode_to(std::FILE *fpout) {
...
+  printf("Otherwise, output is redirected to standard output.\n");
 }
 
 int main(int argc, char *argv[]) {
+  std::string progname = argv[0];
+  // Detect if we are running in GNU mode by checking if the program name
+  // contains "coreutils". Technically, this would allow a program called
+  // potatocoreutilsmywife to trigger GNU mode, but it is a simple heuristic
+  // that should work in practice.
+  std::filesystem::path exe_path(progname);
```
pauldreik, line 503: include filesystem

lemire: @pauldreik I see `#include <filesystem>` in the code? Anythow, it would not build without it, right?

pauldreik: I must have made a mistake, saw it vanish in a commit. It might build anyway, if it is a transitive dependency. nevermind then!

lemire:
@pauldreik The problem was with the python script which I revised.

pauldreik:
I think this is a very nice addition, let's merge!

pauldreik:
The test passes for me now, it is also pretty fast so I see no problem enabling it regardless of the value of SIMDUTF_FAST_TESTS

PR #950, opened 2026-03-25
upgrading cmake

lemire:
Upgrades cmake and use CMAKE_COMPILE_WARNING_AS_ERROR.

### DIFF
--- a/.github/workflows/macos-latest.yml
+++ b/.github/workflows/macos-latest.yml
@@ -17,7 +17,7 @@ jobs:
         run: |
           mkdir build
           cd build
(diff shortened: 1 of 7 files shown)


pauldreik:
is there a particular reason to upgrade cmake?

the mingw ci job failures seem unrelated.

I am for upgrading cmake. But it feels somewhat odd that we hold ourselves back to C++11 while demanding a cmake released 2023.

lemire:
@pauldreik We don't have to upgrade.

> But it feels somewhat odd that we hold ourselves back to C++11 while demanding a cmake released 2023.

This can be revisited. Although, of course, it would require a major release.

I don't think we can require anything beyond C++20, however. Yet C++20 would be fine. We could have two tracks, maintain security patches for the current version, and go C++20 with a new major release.

pauldreik:
I think this sounds like a good proposal - switch to C++20 and a more modern cmake in the next major release, and possibly do security patches on the current release, for those needing to stay on C++11.

PR #957, opened 2026-04-05
Introduce an inlineable API for the 'short inputs' problem

lemire:
Our functions are very fast but they typically cannot be inlined at the calling site (they are not in the headers). This leads to issue https://github.com/simdutf/simdutf/issues/925

How do you solve this generally?

It is tough to do automagically.

So the idea to give the control to the caller. It introduces a new namespace `inlineable` that mimics our main API. So people can call it like so...

```cpp
  if (len < small_input_threshold) {
    return simdutf::inlineable::convert_utf8_to_utf16(input, len, output);
  }
  return simdutf::convert_utf8_to_utf16(input, len, output);
```

Folks can then do the engineering on their side, deciding when to use the inlineable (but slower) function.

WARNING: This PR was done by AI, it is not meant for review per se at this stage. It is purely the expression of an idea. If the idea finds support, the code will be done properly with human review of every line.


lemire:
Note that the tests don't pass nor are they expected to. This is a prototype for illustration purposes only.

pauldreik:
when using the inline api as proposed in this PR, does the code actually get inlined? What I have heard people say is that functions that are more than a few lines long are not so likely to get inlined (without tweaking the compiler inline parameters)

if it doesn't actually get inlined, there is no point diverting to the scalar implementation.

lemire:
> when using the inline api as proposed in this PR, does the code actually get inlined? What I have heard people say is that functions that are more than a few lines long are not so likely to get inlined (without tweaking the compiler inline parameters) if it doesn't actually get inlined, there is no point diverting to the scalar implementation.


Right now, this PR is basically equivalent to adding a single file (there are other changes but they are secondary). So it is not as bad as it looks.

But this is a very good question.

lemire:
So using our short benchmark tool, I raw benchmarks with inputs of 1 element. The most interesting cases are our length computation functions. You see that the inlineable versions run in about 1.5 ns (this includes some overhead, obviously) whereas our conventional functions need over 3 ns, no matter what.

benchmark_length1

The command is

```
./build/benchmarks/shortbench --step 8 --max-size 16 --all
```




At least in my benchmark, the gains are not obvious once we reach 9 elements of inputs.

benchmark_length9


So that's not very conclusive.

lemire:
@pauldreik Maybe there is no hope for this direction.

lemire:
For people reading this: this PR is in draft mode for a reason.

PR #958, opened 2026-04-09
reduce the size of our tools

lemire:
On my mac, I get this...

| Tool       | Before (full simdutf)       | After (feature-specific)    | Reduction    |
|------------|-----------------------------|-----------------------------|--------------|
| fastbase64 | 223 KB (208 KB stripped)    | 131 KB (119 KB stripped)    | 41% smaller  |
| sutf       | 223 KB (203 KB stripped)    | 150 KB (137 KB stripped)    | 32% smaller  |

Interestingly, I get

```
$ which base64
/usr/bin/base64
$ ls -alh /usr/bin/base64
-rwxr-xr-x  5 root  wheel   133K août  16  2025 /usr/bin/base64
```

vs

```
ls -al build/tools/fastbase64
-rwxr-xr-x@ 1 dlemire  staff  133832 avr.   9 15:20 build/tools/fastbase64
```

This is possible through the magic of features (thanks to @WojciechMula for the idea).
So, on my system, `fastbase64` ends up SMALLER.

This makes no sense to me. We should be much fatter.

Update: the trick is that we depend on the C++ lib which is a dynamic library under macOS.


lemire:
I think that copilot is drunk.

pauldreik:
interesting experiment!
on my system (debian trixie, amd64), building with "-O3 -march native", fastbase64 is 1.2M stripped.

If I build like this:
```shell
#!/bin/sh
singleheader/amalgamate.py --with-base64
(cat singleheader/simdutf.h
grep -v '#include "simdutf.h"' singleheader/simdutf.cpp
tail -n +2 tools/fastbase64.cpp ) >main.cpp

/usr/lib/ccache/g++-14 -march=native -O3 -DNDEBUG -std=c++17 main.cpp
```
the stripped result is 119 Kb.

lemire:
@pauldreik Really?

Here is what I get on Rocky Linux (x86)...


Before this PR:

```
465824 build/tools/fastbase64
```

with this PR...

```
204184 build/tools/fastbase64
```

So a dramatic reduction.

On this machine

```
Rocky Linux release 10.1 (Red Quartz)
```

This system assumes `x86_64-v3`. So simdutf just builds the haswell and icelake kernel (as no other kernel is needed).

I think Debian still seek to support everything all the way back to the Pentium 4 so you are going to have 4 kernels. It might be up to 2x larger.  Still, you should see a massive reduction in the size of the binary with this PR.


Can you double check?

pauldreik:
I am sorry, I should have been clearer that I ran on master. I checked the flags used by a default build and repeated the test. Here is my full test script.

```shell
#!/bin/sh
singleheader/amalgamate.py --with-base64
(cat singleheader/simdutf.h
grep -v '#include "simdutf.h"' singleheader/simdutf.cpp
tail -n +2 tools/fastbase64.cpp ) >main.cpp

c++ -O3 -DNDEBUG -std=c++17 -mno-avx256-split-unaligned-load -mno-avx256-split-unaligned-store main.cpp -o fastbase64.onetu
strip fastbase64.onetu
ls -lah fastbase64.onetu # 243K

c++ -O3 -march=native -DNDEBUG -std=c++17 -mno-avx256-split-unaligned-load -mno-avx256-split-unaligned-store main.cpp -o fastbase64.onetu-native
strip fastbase64.onetu-native
ls -lah fastbase64.onetu-native # 119K

cmake -B /tmp/blaha -S .
cmake --build /tmp/blaha/
strip /tmp/blaha/tools/fastbase64
ls -lah /tmp/blaha/tools/fastbase64 # 1.1 M before PR, 772K after
```

So this PR indeed reduces the size (from 1.1M to 772K).

On my machine, the single translation unit gets it down to 243K and to 119K if compiling for the current architecture.

lemire:
Ok. I am merging this since it is beneficial (less bloat).

PR #961, opened 2026-04-17
Remove unused headers

pauldreik:
While reviewing https://github.com/simdutf/simdutf/pull/959 I was surprised seeing iostream included, so I fixed that and also moved inclusion of algorithm to where it is used.


lemire:
@pauldreik `<algorithm>` is a huge header and we should probably get rid of it for good.

I think we should roll our own `find` function.

pauldreik:
closing this in favor of #1010

PR #963, opened 2026-04-18
we should not be using exceptions

lemire:
Exceptions can be disabled and having an exception there will break code.

### DIFF
--- a/include/simdutf/implementation.h
+++ b/include/simdutf/implementation.h
@@ -7274,7 +7274,7 @@ consteval auto base64_decode_literal(const char *str) {
   auto r = scalar::base64::base64_to_binary_details_impl(
       str, InputLen, result.buffer.data(), base64_default, loose);
   if (r.error != error_code::SUCCESS) {
-    1 / 0; // invalid base64 input in _base64 literal
+    (void)(1 / 0); // invalid base64 input in _base64 literal
   }
   result.output_count = r.output_count;
   return result;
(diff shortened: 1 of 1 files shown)


include/simdutf/implementation.h
```diff
@@ -7274,7 +7274,7 @@ consteval auto base64_decode_literal(const char *str) {
   auto r = scalar::base64::base64_to_binary_details_impl(
       str, InputLen, result.buffer.data(), base64_default, loose);
   if (r.error != error_code::SUCCESS) {
-    throw "invalid base64 input in _base64 literal";
+    (void)(1 / 0); // invalid base64 input in _base64 literal
```
pauldreik, line 7277: this causes warnings.

I tried some tricks but everything failed me. the best I can do is to suggest replacing it with std::unreachable() from c++23, which we can use at this particular place.

pauldreik: using this approach:

```cpp
   std::unreachable(); // failed decoding literal!
  }
```

 gcc spits out the following error message when using an invalid base64 literal:

> /home/pauldreik/code/delaktig/simdutf/include/simdutf/implementation.h:7272:20: error: call to non-‘constexpr’ function ‘void std::unreachable()’
 7272 |    std::unreachable(); // failed decoding literal!

lemire:
I am starting to like https://github.com/simdutf/simdutf/pull/962 very much, and it includes this change, so I will drop this PR.

PR #964, opened 2026-04-22
build with fil-c

lemire:
This PR adds support for building simdutf with [Fil-C, the memory-safe C/C++ compiler](https://fil-c.org).

For the most part, we just have to adjust our cpu detection.

There is a tricky issue with masked AVX-512 writes. I think that receiving a null pointer in such cases ought to be safe the mask is empty but Fil-C disagrees. I think that Fil-C is wrong.

cc @pizlonator

### DIFF
--- a/include/simdutf/internal/isadetection.h
+++ b/include/simdutf/internal/isadetection.h
@@ -50,8 +50,12 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <cstdlib>
 #if defined(_MSC_VER)
   #include <intrin.h>
-#elif defined(HAVE_GCC_GET_CPUID) && defined(USE_GCC_GET_CPUID)
+#elif (defined(HAVE_GCC_GET_CPUID) && defined(USE_GCC_GET_CPUID)) || defined(__FILC__)
   #include <cpuid.h>
+#endif
+
+#ifdef __FILC__
+#include <stdfil.h>
 #endif
 
 #include "simdutf/portability.h"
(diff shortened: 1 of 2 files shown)


pauldreik:
It feels somewhat wasteful to add an extra branch, but I assume the performance impact is negligible. 

I made a try with building it in a ubuntu 24.04 docker container and that was no problem. Should I add a CI job on this branch (or if you want to do it)?

The commands I ran were:

```shell
docker run -it  -v $(pwd):/src ubuntu:24.04

apt install wget xz-utils patchelf cmake build-essential
wget 'https://github.com/pizlonator/fil-c/releases/download/v0.678/filc-0.678-linux-x86_64.tar.xz'
tar xf filc-0.678-linux-x86_64.tar.xz
cd cd filc-0.678-linux-x86_64
export CXX=build/bin/clang++ 
export CC=build/bin/clang
cmake -B /tmp/blaha -S /src/
cmake --build /tmp/blaha -j8
ctest --test-dir /tmp/blaha -j8
```

lemire:
@pauldreik It is something like an extra branch per function call (these branches are all in the tail). So, no, it won't be measurable. It should also be a predictable branch...

lemire:
@pauldreik What to add Fil-C to the CI?

As a side benefit, this acts as a sanitizer. Fil-C will crash on unsafe accesses.

pauldreik:
I think it is great to add all ways possible to weed out bugs! 

yes, I would like to add a ci job compiling and testing simdutf with fil-c.

lemire:
I'll merge this for now.

Issue #965, opened 2026-04-25
compile error with clang-22 on debian 11


gvollant:
I'm trying compile simdutf on my debian 11 comptuer, with clang-22 from https://apt.llvm.org

```
gillesvollant@gv-w11x64:/mnt/w/devw$ clang++-22 -std=c++23 -o vir exterieur/simdutf/singleheader/simdutf.cpp
In file included from exterieur/simdutf/singleheader/simdutf.cpp:3:
exterieur/simdutf/singleheader/simdutf.h:14111:10: error: no member named 'unreachable' in namespace 'std'
 14111 |     std::unreachable(); // invalid base64 input in _base64 literal
       |          ^~~~~~~~~~~
exterieur/simdutf/singleheader/simdutf.cpp:33262:32: warning: array index 2 is past the end of the array (that has type 'const simd8<unsigned char>[2]') [-Warray-bounds]
 33262 |         this->check_utf8_bytes(input.chunks[2], input.chunks[1]);
       |                                ^            ~
exterieur/simdutf/singleheader/simdutf.cpp:4826:3: note: array 'chunks' declared here
 4826 |   simd8<T> chunks[NUM_CHUNKS];
      |   ^
exterieur/simdutf/singleheader/simdutf.cpp:33263:32: warning: array index 3 is past the end of the array (that has type 'const simd8<unsigned char>[2]') [-Warray-bounds]
 33263 |         this->check_utf8_bytes(input.chunks[3], input.chunks[2]);
       |                                ^            ~
exterieur/simdutf/singleheader/simdutf.cpp:4826:3: note: array 'chunks' declared here
 4826 |   simd8<T> chunks[NUM_CHUNKS];
      |   ^
exterieur/simdutf/singleheader/simdutf.cpp:33263:49: warning: array index 2 is past the end of the array (that has type 'const simd8<unsigned char>[2]') [-Warray-bounds]
 33263 |         this->check_utf8_bytes(input.chunks[3], input.chunks[2]);
       |                                                 ^            ~
exterieur/simdutf/singleheader/simdutf.cpp:4826:3: note: array 'chunks' declared here
 4826 |   simd8<T> chunks[NUM_CHUNKS];
      |   ^
3 warnings and 1 error generated.
```

I've no error with  --std=c++20

pauldreik:
[...]

PR #966, opened 2026-04-25
workaround missing std::unreachable

pauldreik:
This adds a workaround for using C++23 and a too old standard library. See https://github.com/simdutf/simdutf/issues/965

### DIFF
--- a/include/simdutf/implementation.h
+++ b/include/simdutf/implementation.h
@@ -7268,7 +7268,12 @@ consteval auto base64_decode_literal(const char *str) {
   auto r = scalar::base64::base64_to_binary_details_impl(
       str, InputLen, result.buffer.data(), base64_default, loose);
   if (r.error != error_code::SUCCESS) {
+  #if __cpp_lib_unreachable >= 202202L
     std::unreachable(); // invalid base64 input in _base64 literal
+  #else
+    // workaround for older stdlib
+    throw "invalid base64 input in _base64 literal";
+  #endif
   }
   result.output_count = r.output_count;
   return result;


pauldreik:
@lemire  this solves the problem according to gvollant. if we should merge it, that is another question.

gvollant:
> @lemire this solves the problem according to gvollant. if we should merge it, that is another question.

yes, thank you a lot for this fix!

include/simdutf/implementation.h
```diff
@@ -7268,7 +7268,12 @@ consteval auto base64_decode_literal(const char *str) {
   auto r = scalar::base64::base64_to_binary_details_impl(
       str, InputLen, result.buffer.data(), base64_default, loose);
   if (r.error != error_code::SUCCESS) {
+  #if __cpp_lib_unreachable >= 202202L
```
lemire, line 7271: This is fine. In similar cases, we created a macro, shortly after including `<version>`. So we could do that. Like  `SIMDUTF_USE_UNREACHABLE` which default to 1 if `__cpp_lib_unreachable >= 202202L`.

The benefit of the macro is that we hand back some control to the user: they can prefer an exception (for whatever reason).

This is nitpicking and your PR is fine.

lemire (approved):

include/simdutf/implementation.h
pauldreik: In general I agree with you about the macro, but in this case it always happens at consteval time so I do not think it matters - it is just there to terminate the consteval evaluation.

PR #970, opened 2026-05-05
Remove redundant checks in scalar utf8 validation

### LINKED ISSUE #969: Remove two redundant expressions in scalar UTF8 validation
During experimentation, I was playing around with the scalar utf8 validation code and identified two expressions that by necessity always evaluate to the same value (`false`).

https://github.com/simdutf/simdutf/blob/427a6f525f5051d448f9337220211ec21332f542/include/simdutf/scalar/utf8.h#L56-L60
```c++
code_point = (byte & 0b00011111) << 6 | (data[pos + 1] & 0b00111111);
if ((code_point < 0x80) || (0x7ff < code_point)) {
    return false;
}
```
- `code_point` is a `uint32_t`
- `byte` is a `unsigned char`
- `data` is a `unsigned char*` (Note: this might not be entirely true. It's definitely something comparable, but I don't believe it makes a difference)

The way `code_point` is evaluated, only the 11 least significant bits can be active.
`0x7ff` is the number with exactly the 11 least significant bits active.
So the expression `0x7ff < code_point` always evaluates to `false`, as at least the twelfth bit would need to be active which isn't possible.

The same goes for the later check https://github.com/simdutf/simdutf/blob/427a6f525f5051d448f9337220211ec21332f542/include/simdutf/scalar/utf8.h#L72-L79 with the expression `0xffff < code_point`.

Here is the relevant code in the compiler explorer: https://godbolt.org/z/9r4fhqW7E.
Note that removing either check reduces the amount of instructions produced, meaning the compiler did not catch this optimization (on -O3).

That also made me skeptical, but the compiler indeed optimizes those checks out in a simpler example: https://godbolt.org/z/M135ejWea.

These two checks each appear twice in the following files:
```
include/simdutf/scalar/utf8_to_utf16/utf8_to_utf16.h
include/simdutf/scalar/utf8_to_utf32/utf8_to_utf32.h
include/simdutf/scalar/utf8.h
```
Also in `tests/reference/validate_utf8.cpp`, but as a reference implementation, it should probably be left as-is.

These two checks also exists in ports of this library, e.g. https://github.com/simdutf/SimdUnicode/blob/2adc753b41cdc167836dc64c2af133368de19b1b/src/UTF8.cs#L125-L130.

If I didn't miss anything, I'll happily follow up with a pull request :)

BenjaminBucher:
The prior bit masks and shifts ensure that the checks always evaluate to `false`. The optimization wasn't caught by the compiler. Fixes #969.

### DIFF
--- a/include/simdutf/scalar/utf8.h
+++ b/include/simdutf/scalar/utf8.h
@@ -55,7 +55,7 @@ simdutf_constexpr23 simdutf_warn_unused bool validate(BytePtr data,
       }
       // range check
       code_point = (byte & 0b00011111) << 6 | (data[pos + 1] & 0b00111111);
-      if ((code_point < 0x80) || (0x7ff < code_point)) {
+      if (code_point < 0x80) {
         return false;
       }
     } else if ((byte & 0b11110000) == 0b11100000) {
@@ -73,7 +73,7 @@ simdutf_constexpr23 simdutf_warn_unused bool validate(BytePtr data,
       code_point = (byte & 0b00001111) << 12 |
                    (data[pos + 1] & 0b00111111) << 6 |
                    (data[pos + 2] & 0b00111111);
-      if ((code_point < 0x800) || (0xffff < code_point) ||
+      if ((code_point < 0x800) ||
           (0xd7ff < code_point && code_point < 0xe000)) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


BenjaminBucher:
(nevermind, got confused by a PR in another project)

lemire:
Excellent. Let us run test tests.

(For people who noticed the confusion... I simply ported these changes to a C# lib in this org called SimdUnicode.)

pauldreik:
is there anything that stops this from being merged?

lemire:
Merging.

PR #971, opened 2026-05-06
scripts/add_function.py: fixes

BenjaminBucher:
The script add_function.py didn't work properly for me, and compiling the generated code produced various errors and warnings.

Fixes I did:
- More precise #endif locating (it got confused by nested #ifs)
- Fix 'const' function specifier detection (it was satisfied by 'const' in a parameter type)
- Uncomment important functionality

The example from the `scripts/README_ADD_FUNCTION.md` now runs and compiles fine with the fixes, as long as you avoid a naming conflict by slightly modifying the names, e.g.:
```
#if SIMDUTF_FEATURE_UTF8 && SIMDUTF_FEATURE_UTF16
/**
 * Documentation for first function.
 */
simdutf_warn_unused size_t utf8_length_from_utf16le_abc(
    const char16_t *buf, size_t len) noexcept;

/**
 * Documentation for second function.
 */
simdutf_warn_unused size_t utf8_length_from_utf16be_abc(
    const char16_t *buf, size_t len) noexcept;
#endif
```

### DIFF
--- a/scripts/add_function.py
+++ b/scripts/add_function.py
@@ -138,7 +138,7 @@ def add_to_src_implementation_cpp(repo_root, feature_macro, functions):
     if detect_start != -1:
         existing_if = content.find(f'#if {feature_macro}', detect_start)
         if existing_if != -1:
-            endif_pos = content.find('#endif', existing_if)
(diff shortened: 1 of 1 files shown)


lemire:
Merging!

PR #973, opened 2026-05-06
add ci job compiling with gcc 16

pauldreik:


lemire:
Oh Oh Oh I did not know we could do this yet !!!!!!!!!!

❤️

lemire:
We could compile with C++26!

pauldreik:
we already have a gcc 15 C++-26 job defined in the debian-latestcxxstandards.yml job !

once debian forky has updated their version of gcc 16 to the released one instead of a snapshot, we can update that job to use gcc 16 instead of 15.

PR #974, opened 2026-05-12
draft: Port `GetPointerToFirstInvalidByte` from SimdUnicode

BenjaminBucher:
**This PR is a draft, so it is very much WIP but open to feedback**

This is a step towards #147 for transcoding UTF8 to UTF16 with replacement.
The new function locates the first invalid byte of UTF8 and also gathers information relevant to transcode to UTF16.

SimdUnicode
---
The function is already implemented for some platforms in C# over at [SimdUnicode](https://github.com/simdutf/SimdUnicode).
In this PR I port that function to simdutf, called `GetPointerToFirstInvalidByte` in SimdUnicode.

In SimdUnicode, the function returns the pointer to the next invalid byte of UTF8, the number of utf16 code units and the amount of characters/code points in total.
The number of utf16 units is required to calculate the output length.
Does the amount of total characters help in any way? I could imagine some niche optimization relying on `number of utf16 units == number of total characters` but am unsure if this is a thing.

Use in Transcoding UTF8 to UTF16
---

Transcoding UTF8 to UTF16 could be done in 2 passes using this function.
Note that the idea is to assume little to no encoding errors. We determine a reasonable threshold N beforehand for how many errors we expect at most.
If the number of actual errors is higher, the transcoding will be significantly slower as it will have to once again have to scan for errors on the second pass.

Step 1: Determine UTF16 length and error locations:
- Determine input slices of valid and invalid UTF8
- Determine output UTF16 length (with replacement logic)
- Store info for up to N valid ranges

Step 1.5: User needs to find / allocate sufficient space.
Step 2: Transcoding with replacement
- Handle stored valid slices with existing transcoding functions, invalid slices with additional logic
- In the case more than N errors, the valid/invalid slices must be rediscovered on the fly

Open questions
---

- Name of the function
- Should this function even appear in the public API
- What extra info should the function return?
[x] Amount of UTF16 units
[ ] Amount of characters
[ ] Correct UTF8 error code
[ ] More?

I'm now hopping into porting the actual SIMD part of the code, wish me luck :)


validate_utf8_for_utf16_transcoding.sig
```diff
@@ -0,0 +1,14 @@
+#if SIMDUTF_FEATURE_UTF8 && SIMDUTF_FEATURE_UTF16
```
lemire: I think you don't want to include this file?

lemire:
@BenjaminBucher 

Thanks for the PR. What about making the function more generic ?

Here is a gist (not a proposal)...


```cpp
struct utf8_segment {
  size_t length; // length in bytes
  size_t ascii; // number of ASCII characters
  size_t two_bytes; // number of two-byte (UTF-8) characters
  size_t three_bytes; // number of three-byte (UTF-8) characters
  size_t four_bytes; // number of four-byte (UTF-8) characters
}
```

And then we have something like...

```cpp
utf8_segment  find_first_invalid_utf8(const char * input, size_t length);
```


This would be a great building block that could be used for various applications, not just UTF-8 to UTF-16 transcoding.

cc @pauldreik @anonrig @erikcorry

validate_utf8_for_utf16_transcoding.sig
BenjaminBucher: Huh, I thought I already replied here already.
I wanted to keep this file in the tree as I'll probably still change the signature a couple of times. Will remove and squash eventually

BenjaminBucher:
I'm definitely in favor of making this function more generic.
The concrete error code could also be returned by the function, as it is quite cheap to compute if we already compute the error location (and might also be helpful).

lemire:
@BenjaminBucher Yeah.


What we want is the common case to be fast (which is going to be proper UTF-8).

BenjaminBucher:
I'm highly confused by the current compile error:
```
In file included from /src/src/haswell/implementation.cpp:103,
                 from /src/src/simdutf.cpp:133:
/src/src/generic/ascii_validation.h: In function 'simdutf::result simdutf::simdutf::haswell::{anonymous}::ascii_validation::generic_validate_ascii_with_errors(const char*, size_t)':
/src/src/generic/ascii_validation.h:7:3: error: 'buf_block_reader' was not declared in this scope; did you mean 'simdutf::haswell::{anonymous}::buf_block_reader'?
    7 |   buf_block_reader<64> reader(reinterpret_cast<const uint8_t *>(input), length);
      |   ^~~~~~~~~~~~~~~~
      |   simdutf::haswell::{anonymous}::buf_block_reader
```
It complains about `buf_block_reader` not being available in `ascii_validation.h`.
I do not understand which of my changes might've broken the compilation of that file. `buf_block_reader` *should* be perfectly available and is used in the same way in `utf8_validator.h`, which is a file I did modify.

pauldreik:
> I'm highly confused by the current compile error:
> 
> ```
> In file included from /src/src/haswell/implementation.cpp:103,
>                  from /src/src/simdutf.cpp:133:
> /src/src/generic/ascii_validation.h: In function 'simdutf::result simdutf::simdutf::haswell::{anonymous}::ascii_validation::generic_validate_ascii_with_errors(const char*, size_t)':
> /src/src/generic/ascii_validation.h:7:3: error: 'buf_block_reader' was not declared in this scope; did you mean 'simdutf::haswell::{anonymous}::buf_block_reader'?
>     7 |   buf_block_reader<64> reader(reinterpret_cast<const uint8_t *>(input), length);
>       |   ^~~~~~~~~~~~~~~~
>       |   simdutf::haswell::{anonymous}::buf_block_reader
> ```
> 
> It complains about `buf_block_reader` not being available in `ascii_validation.h`. I do not understand which of my changes might've broken the compilation of that file. `buf_block_reader` _should_ be perfectly available and is used in the same way in `utf8_validator.h`, which is a file I did modify.

sometimes it is helpful to do an amalgamation, and work with the amalgamated source. the include system can be very confusing.

src/generic/utf8_validation/utf8_validator.h
```diff
@@ -78,6 +78,68 @@ result generic_validate_utf8_with_errors(const char *input, size_t length) {
...
+    return utf8_result(error_code::SUCCESS, length, c.continuation_count(),
+                       c.four_byte_count());
+  }
+}
+
+utf8_result generic_validate_utf8_with_counts(const char *input,
+                                              size_t length) {
+  return generic_validate_utf8_with_counts<utf8_segmenter>(
+      reinterpret_cast<const uint8_t *>(input), length);
+
```
lemire, line 133:
```suggestion
}
```

You never seem close the scope of the function. That's going to give troublesome error messages.

src/generic/utf8_validation/utf8_lookup4_algorithm.h
```diff
@@ -214,9 +229,76 @@ struct utf8_checker {
...
+    std::pair<size_t, size_t> cont_4byte = count_cont_4byte(input);
+    this->continuations += cont_4byte.first;
+    this->four_byte += cont_4byte.second;
+    this->checker.check_utf8_bytes(input, prev_input);
+  }
+
+  simdutf_really_inline void check_eof() { this->checker.check_eof(); }
+
+  simdutf_really_inline std::pair<size_t, size_t>
+  check_next_input_with_counts(const simd8x64<uint8_t> &input) {
```
lemire, line 270:
```suggestion
  check_next_input(const simd8x64<uint8_t> &input) {
```

lemire:
@pauldreik 

> sometimes it is helpful to do an amalgamation, and work with the amalgamated source. the include system can be very confusing.

Ah Ah

I think it is quite simple and a classic: a missing final brace (`}`) which makes the C++ compiler go wild and report strange errors.

src/generic/utf8_validation/utf8_lookup4_algorithm.h
lemire: I think check_next_input_with_counts is never called, but check_next_input is.

src/generic/utf8_validation/utf8_validator.h
BenjaminBucher: Nice spot, I have to admit the compiler error threw me off guard. Will keep that in mind as a sanity check if I ever encounter some very unexpected compiler error ^^

BenjaminBucher:
@pauldreik Yes, the include system did confuse me a couple of times already, but I'm slowly getting a feel for it. Thanks for the tip with the amalgamation.
I also kind of gave up using an IDE for now an am just working with the compiler messages the classic way :)

include/simdutf/scalar/utf8.h
```diff
@@ -395,7 +395,7 @@ inline simdutf_warn_unused utf8_result rewind_and_validate_with_counts(
     extra_len++;
     if (byte >= 0b11110000) {
       extra_four_bytes += 1;
-    } else if ((char)byte < -65 + 1) {
+    } else if ((signed char)byte < -65 + 1) {
```
lemire: this is fine, simpler might be `int8_t`.

pauldreik:
another tip for detecting silly mistakes like a missing brace is to run the clang format script: you often get large changes in the wrong place.

see scripts/clang_format.sh

BenjaminBucher:
@lemire 
I fear that the linear system of equations to calculate number of ascii symbols and 2-/3-byte leads with the current set of counters is underdetermined.

We got: length (amount of bytes), ncon (amount of continuation bytes), n4 (amount of 4-byte leads)
We want: n1, n2, n3 (amount of ascii bytes, 2-/3-byte leads)
```
length = n4 * 4 + n3 * 3 + n2 * 2 + n1
ncon = n4 * 3 + n3 * 2 + n2
```
If I didn't miss anything, we only have 2 equations.
To calculate the utf16 length we can luckily do without solving the linear equation.
However, as we have fewer equations than unknowns, we can't provide all these numbers with just keeping track of the amount of continuation bytes and 4-byte leads.

We'd need to keep track of another variable, which we can do, but will add more runtime overhead not required for utf16 length.
This might not be too bad, but a relevant decision, which is why I want input :)

lemire:
@BenjaminBucher You are, of course, correct.

I think we need at least one more piece of information if we want to proceed with my sketch of an idea: the total number of ASCII bytes ($n_1$).


Let \(L\) be the total number of bytes in the string. We define:

- $n_1$: number of 1-byte characters (ASCII)
- $n_2$: number of 2-byte characters  
- $n_3$: number of 3-byte characters
- $n_4$: number of 4-byte characters
- $n_{\rm con}$: number of continuation bytes

We have two expressions for the total byte length $L$:

$$
L = 4n_4 + 3n_3 + 2n_2 + n_1
$$

$$
L = n_4 + n_3 + n_2 + n_{\rm con} + n_1
$$


Solving the linear system for $n_3$ and $n_2$:

$$
n_3 = n_1 - 2n_4 + 2n_{\rm con} - L
$$

$$
n_2 = -2n_1 + n_4 - 3n_{\rm con} + 2L
$$



You might object that this will entail extra processing. It is a slight additional cost.

Right.

So maybe we can do something else instead.

What about this:

```cpp
struct utf8_segment {
  size_t length; // length in bytes
  size_t characters; // number of characters ASCII, two-byte, three-byte and 4-byte
  size_t four_bytes; // number of four-byte (UTF-8) characters
}
```

The number of characters is:

$$
N = n_1 + n_2 + n_3 + n_4 = L - n_{\rm con}
$$

That's it. Characters = bytes minus continuation bytes.

This generic data structure is enough to compute the required UTF-16 output (it is just `characters + four-bytes`). But also the required UTF-32 output (it is just `characters`). 


The benefit of my approach, I think, it is that it is easier to debug.

The downside is that the data structure is slightly fatter than what is needed, if we just want to go to UTF-16, then we have slightly too much information. So there is a tiny performance penalty here, with my proposal, but I think that it is so small as to be irrelevant in practice.

BenjaminBucher:
With the focus on throughput, I first definitely recoiled from the idea of keeping yet another counter.
However, I am easily convinced that the actual impact can be negligible.

Keeping track of ascii/non-ascii characters sounds simple enough.

I'll adjust the generic implementation. In general, it could be slightly improved by duplicating code more code and integrating the counters deeper in the simd functions.
Next I'll take a look at the avx implementation :)

lemire:
@BenjaminBucher I am not arguing that we need to keep track of the ASCII characters. It may indeed be quite cheap.

BenjaminBucher:
Fair fair, that sounds to me like a good solution :)

BenjaminBucher:
Among other things the compilers appear to not particularly like the tuples. They aren't that great anyways and should probably be replaced with a struct. Can I define the struct somewhere where it won't appear in the public API?

lemire:
@BenjaminBucher I think you are doing fine, the errors look like details we can work out later.

This being said, I agree that a dedicated struct is better. Why not make it public? There is nothing wrong with public structures?

BenjaminBucher:
Fair point. I'm typically in favor of keeping the public interface minimal, and I wasn't convinced that the benefits of a struct for this are sufficient :)

BenjaminBucher:
The icelake implementation is embarrassingly simple and I want to propose to do the same in the generic implementation.
In the generic implementation I tried to stay close to the simdunicode implementation and because of that dug *relatively* deep into the simd validation code. I still tried to keep the validation logic somewhat separate to counts, in order to keep the logic simple and easier to verify.

For the icelake implementation I didn't check the simdunicode implementation and found it significantly easier to keep the counts out of the validator and instead do the counts where the reader is processed, chunk for chunk.

Trade-offs (non-exhaustive):
- Doing the counts alongside the validation logic would save ~2 avx-512 comparison operations, maybe another load but for that I'd need to look into the assembly. However, this would require us to duplicate the entire validation code and changing it in a few places with more state, which would make the code more complex.
- In the generic code, one reader chunk contains either 2 or 4 full simd registers worth of data. The count logic is currently inserted in a location where each of the simd register parts of the chunk is loaded only once. (Once again, at the cost of more code complexity).

I don't have good intuition for where overoptimization starts and where good-enough and simple code is in order. Feedback wanted :)

lemire:
I’ll get back to you in the next few days.

validate_utf8_with_counts.sig
lemire, line 1: I think that we don't want this file checked in.

lemire: Sorry, I see that you want to keep it temporarily.

lemire:
@BenjaminBucher Sorry for the delay.

Ok. At this stage, we don't care about performance very much, at least not down to the details of each implementation. What we care a great deal is to get a correct results.

src/icelake/implementation.cpp
```diff
@@ -242,6 +242,60 @@ simdutf_warn_unused result implementation::validate_utf8_with_errors(
...
+      res.four_byte_count += four_byte_leads;
+      res.non_ascii_count += non_ascii;
+      return res;
+    }
+    continuations += utf8_count_continuations(utf8);
+    four_byte_leads += utf8_count_4_byte_leads(utf8);
+    non_ascii += utf8_count_non_ascii(utf8);
+    count += 64;
+  }
+  const __m512i utf8 = _mm512_maskz_loadu_epi8(
```
lemire: @BenjaminBucher 

If end == ptr, we have `~UINT64_C(0) >> 64` which compiles as  `~UINT64_C(0) >> 0` (it is undefined).

So this is not safe.

Because you branch on `end!=ptr` anyhow,  you should guard this code.

src/generic/utf8_validation/utf8_validator.h
```diff
@@ -78,6 +78,73 @@ result generic_validate_utf8_with_errors(const char *input, size_t length) {
...
+                                              size_t length) {
+  checker c{};
+  buf_block_reader<64> reader(input, length);
+  size_t count{0};
+  while (reader.has_full_block()) {
+    simd::simd8x64<uint8_t> in(reader.full_block());
+    block_counts last_counts = c.check_next_input_with_counts(in);
+    if (c.errors()) {
+      /*
+       * Why was this done?
```
lemire: You have ` rewind_and_validate_with_counts` which handles this part.

include/simdutf/scalar/utf8.h
```diff
@@ -248,6 +248,184 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
...
+// credit: based on code from Google Fuchsia (Apache Licensed)
+// TODO: upon finalizing, figure out constexpr23 stuff?
+// TODO: also only-ascii optimization?
+template <class BytePtr>
+simdutf_constexpr23 simdutf_warn_unused utf8_result
+validate_utf8_with_counts(BytePtr data, size_t len) noexcept {
+  static_assert(
+      std::is_same<typename std::decay<decltype(*data)>::type, uint8_t>::value,
+      "dereferencing the data pointer must result in a uint8_t");
+  uint64_t pos = 0;
```
lemire:
```suggestion
  size_t pos = 0;
```

include/simdutf/scalar/utf8.h
```diff
@@ -248,6 +248,184 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
...
+  uint32_t code_point = 0;
+  uint64_t continuations = 0;
```
lemire:
```suggestion
  size_t continuations = 0;
```

include/simdutf/scalar/utf8.h
```diff
@@ -248,6 +248,184 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
...
+  uint64_t four_byte = 0;
```
lemire:
```suggestion
  size_t four_byte = 0;
```

include/simdutf/scalar/utf8.h
```diff
@@ -248,6 +248,184 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
...
+  uint64_t non_ascii = 0;
```
lemire:
```suggestion
  size_t non_ascii = 0;
```

include/simdutf/scalar/utf8.h
```diff
@@ -248,6 +248,184 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
...
+// TODO: upon finalizing, figure out constexpr23 stuff?
+// TODO: also only-ascii optimization?
+template <class BytePtr>
+simdutf_constexpr23 simdutf_warn_unused utf8_result
+validate_utf8_with_counts(BytePtr data, size_t len) noexcept {
+  static_assert(
+      std::is_same<typename std::decay<decltype(*data)>::type, uint8_t>::value,
+      "dereferencing the data pointer must result in a uint8_t");
+  uint64_t pos = 0;
+  uint32_t code_point = 0;
```
lemire, line 262: The type code be size_t ?

include/simdutf/scalar/utf8.h
```diff
@@ -248,6 +248,184 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
...
+  uint64_t continuations = 0;
+  uint64_t four_byte = 0;
+  uint64_t non_ascii = 0;
+  while (pos < len) {
+    uint64_t next_pos;
```
lemire:
```suggestion
    size_t next_pos;
```

lemire:
@BenjaminBucher The AVX-512 is not quite correct, see my comment. Once this is fixed, then I expect your code to be correct.

The next step would then be to wire in this function into our benchmarks, and, then, once we have benchmarks, it is easier to take decisions regarding the code.

As a rule, have slightly more complicated code for better performance is a good tradeoff in simdutf. Our main selling point is performance, not code simplicity.

BenjaminBucher:
I can hardly call a single day a delay ^^.
Where is your comment on the avx512 code, I'm not sure whether it was sent.
I'll take a look at benchmarking.

lemire:
@BenjaminBucher You should now see my comments. They were in the system but not yet visible.

validate_utf8_with_counts.sig
BenjaminBucher: I'll have it in-tree until I rebase/squash anyway, so I removed it now

src/icelake/implementation.cpp
BenjaminBucher: Good catch. The control flow is a little odd there, as we needed that utf8 segment in case there's no error later on to continue the counters. Switched it up to update a second set of counters so we have either for the different code paths.

lemire:
@BenjaminBucher Please consider this PR https://github.com/BenjaminBucher/simdutf/pull/2 merging it would change this PR.

I think we are quite close to finishing off this PR.

lemire:
@BenjaminBucher Please see also https://github.com/BenjaminBucher/simdutf/pull/3

lemire:
@BenjaminBucher I have a few pull request to your branch. I think we want this new functionality to be public. We are going to build on it, but I think it is already useful.

BenjaminBucher:
^ rebased onto master to fix conflicts

src/generic/utf8_validation/utf8_lookup4_algorithm.h
```diff
@@ -214,9 +255,76 @@ struct utf8_checker {
...
+      return block_counts{0, 0};
+    } else {
+      size_t prev_continuations = this->continuations;
+      size_t prev_four_byte = this->four_byte;
+      // you might think that a for-loop would work, but under Visual Studio, it
+      // is not good enough.
+      static_assert((simd8x64<uint8_t>::NUM_CHUNKS == 2) ||
+                        (simd8x64<uint8_t>::NUM_CHUNKS == 4),
+                    "We support either two or four chunks per 64-byte block.");
+      if (simd8x64<uint8_t>::NUM_CHUNKS == 2) {
```
lemire: **[suggestion]** `utf8_segmenter::check_next_input_with_counts` uses plain `if (NUM_CHUNKS == 2/4)` instead of `if constexpr`, while `utf8_checker::check_next_input` and the rest of the tree (utf8_to_utf16/32, etc.) use `if constexpr` for the same split. On Haswell, `NUM_CHUNKS == 2`, so the `== 4` branch still has to be type-checked and names `input.chunks[2]` / `chunks[3]` on a 2-element array. That is dead at runtime but is exactly why the existing code uses `if constexpr` (and the nearby comment about Visual Studio).

**Suggestion:** Mirror `utf8_checker` and switch both arms to `if constexpr` / `else if constexpr`.

include/simdutf/implementation.h
```diff
@@ -313,6 +313,29 @@ validate_utf8_with_errors(
...
+ *
+ * These two counts are enough to derive, without a second pass over the input,
+ * both the number of code points (input_count - continuation_count) and the
+ * number of UTF-16 code units (input_count - continuation_count +
+ * four_byte_count) of the valid prefix. The utf16_length() helper returns the
+ * latter.
+ *
+ * @param buf the UTF-8 string to validate.
+ * @param len the length of the string in bytes.
+ * @return a utf8_result struct with an error code, the number of bytes in the
```
lemire, line 331: **[suggestion]** The free function `validate_utf8_with_counts(const char*, size_t)` is added without a `SIMDUTF_SPAN` overload, while `validate_utf8` / `validate_utf8_with_errors` (immediately above) expose span/byte-like overloads under `#if SIMDUTF_SPAN`, including consteval scalar paths on C++23. Callers using the modern span API will hit an inconsistent surface for this new entry point.

**Suggestion:** Add a span overload parallel to `validate_utf8_with_errors`, dispatching to the char* API at runtime and to `scalar::utf8::validate_utf8_with_counts` under `if consteval` when applicable.

src/generic/utf8_validation/utf8_lookup4_algorithm.h
```diff
@@ -153,6 +153,43 @@ simdutf_really_inline simd8<uint8_t> is_incomplete(const simd8<uint8_t> input) {
...
+  const simd8<uint8_t> one = simd8<uint8_t>::splat(1);
+  size_t continuations =
+      (simd8<uint8_t>((simd8<int8_t>)input < mask_lt) & one).sum_bytes();
+  size_t four_byte = (simd8<uint8_t>(input >= mask_gte) & one).sum_bytes();
+  return block_counts{continuations, four_byte};
+#else
+  // On x86 a movemask + popcount is the cheapest way to count matching lanes.
+  uint64_t continuation_mask = ((simd8<int8_t>)input < mask_lt).to_bitmask();
+  size_t continuations = count_ones(continuation_mask);
+  int64_t four_byte_mask = (input >= mask_gte).to_bitmask();
```
lemire: **[nit]** `four_byte_mask` is stored as `int64_t` while `continuation_mask` is `uint64_t` and `count_ones` takes `uint64_t`. Harmless for current bitmask widths, but inconsistent (and the same footgun pattern exists in `utf16_length_from_utf8`; no need to widen scope, but new code should not copy the weaker form).

**Suggestion:** Use `uint64_t four_byte_mask` for both masks.

include/simdutf/scalar/utf8.h
```diff
@@ -251,6 +251,171 @@ inline simdutf_warn_unused result rewind_and_validate_with_errors(
   return res;
 }
 
+// credit: based on code from Google Fuchsia (Apache Licensed)
+// TODO: upon finalizing, figure out constexpr23 stuff?
```
lemire: **[nit]** Scalar `validate_utf8_with_counts` still carries TODOs (“constexpr23 stuff?”, “only-ascii optimization?”) and omits the 16-byte ASCII fast path that `validate_with_errors` uses. Fine for a draft, but fallback/RVV pay for it on large ASCII-heavy input.

**Suggestion:** Before finalize: drop or resolve TODOs; consider reusing the same ASCII skip as `validate_with_errors` for parity.

lemire (commented):
A few useful comments...

src/generic/utf8_validation/utf8_lookup4_algorithm.h
BenjaminBucher: curious that they got lost, thanks

include/simdutf/implementation.h
BenjaminBucher: curious that they got lost

src/generic/utf8_validation/utf8_lookup4_algorithm.h
BenjaminBucher: good spot, done

include/simdutf/scalar/utf8.h
BenjaminBucher: As far as I could tell, the constexpr23 comment was outdated. I did the fast path for ASCII input.

src/generic/utf8_validation/utf8_lookup4_algorithm.h
BenjaminBucher: nicely spotted, fixed

include/simdutf/implementation.h
BenjaminBucher: While doing so, I noticed that `scalar::utf8::validate_utf8_with_counts` didn't fit into the naming scheme. Renamed to `scalar::utf8::validate_with_counts`.

lemire:
@BenjaminBucher Can you have a look at the PRs I have opened on your branch ? https://github.com/BenjaminBucher/simdutf/pulls

lemire:
@BenjaminBucher I think that if we merge https://github.com/BenjaminBucher/simdutf/pull/11 then we should mark this as ready for review and then merge it soon after.

lemire:
The typo check is a false positive.

@BenjaminBucher I would merge this as soon as you give the ok. This is good work. If we have that, then we will be able to finish the transcoding with replacement.

lemire:
@BenjaminBucher  I am marking as 'ready for review'.

lemire:
Requested a copilot review.

include/simdutf/implementation.h
```diff
@@ -5316,6 +5355,21 @@ class implementation {
    */
   simdutf_warn_unused virtual result
   validate_utf8_with_errors(const char *buf, size_t len) const noexcept = 0;
+  /**
+   * Finds the pointer to the first byte of invalid utf8 while counting
+   continuation bytes and four-byte sequences.
+
+   * @param buf the UTF-8 string to validate.
+   * @param len the length of the string in bytes.
```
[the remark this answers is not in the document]
lemire: This is nitpicking.

lemire:
@pauldreik @BenjaminBucher I am planning to merge this soon. Please raise objections if you have any.

PR #975, opened 2026-05-19
add test where rewind causes integer underflow

BenjaminBucher:
The rewind function walks back to the latest lead byte and then scans for the next error. If The error is caused in continuation characters of that lead byte, the error lies left of the *buf parameter passed. The function indicates that by returning an underflowed count.

Note that many other algorithms have other rewind functions which I did not test or document.

This behavior was surprising to me, so I thought it would be good to test/document it.

Type of change
- [ ] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation / tests


lemire:
Merged.

PR #978, opened 2026-05-25
fix doubled length in binary_length_from_base64 for char16 input

metsw24-max:
The char16 `binary_length_from_base64` in src/generic/base64lengths.h counts a bitmask that sets two bits per 16-bit lane, so the vectorized loop double-counts and 64 base64 units report 96 bytes instead of 48. Halve the vectorized count before the scalar tail, as the avx2 and icelake kernels already do.


pauldreik:
do I understand this correct - the added test would not succed without the change in `src/generic/base64lengths.h` ?

lemire:
@pauldreik I verified locally that the new test DOES NOT pass. So this is a genuine fix.

pauldreik:
This identifies a hole in the fuzzer coverage! I added std::abort() to the place and ran the fuzzers without hitting it.

PR #985, opened 2026-06-01
scripts/add_function.py: Better #if detection

BenjaminBucher:
Previously, the #if block detection sometimes chose the wrong block if a substring of its condition matched the search string.
Now it also matches the newline to avoid that.

E.g. `#if SIMDUTF_FEATURE_UTF8` is a substring of e.g. `#if SIMDUTF_FEATURE_UTF8 && SIMDUTF_FEATURE_UTF32`.

Type of change
- [x] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests


lemire:
Merged

PR #926, opened 2026-01-23
optimise simdutf for short strings

sleepingeight:
Related to https://github.com/simdutf/simdutf/issues/925

Started with a straightforward implementation, modifying the public API's defination in this way, 
```cpp
simdutf_warn_unused bool validate_utf8(const char *buf, size_t len) noexcept {
  if (len < SIMDUTF_SHORT_INPUT_THRESH) {
    return scalar::utf8::validate(buf, len);
  }
  return get_default_implementation()->validate_utf8(buf, len);
}
```
Which was before,
https://github.com/simdutf/simdutf/blob/bcfbae9fad23aa5c0f893a72ed013416a8567608/src/implementation.cpp#L1429-L1431

Using the `benchmark` utiltiy, got 0 improvements. But when benchmarked using google-benchmark for input sizes from 1 to 15, these are the results for `validate_utf8` procedure - 

Old: 
```
Running ./bench_utf8
Run on (10 X 24 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x10)
Load Average: 2.50, 2.81, 2.65
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
BM_validate_utf8_small/1      840503 ns       839682 ns          755 bytes_per_second=93.0412Mi/s
BM_validate_utf8_small/2      846647 ns       842714 ns          830 bytes_per_second=185.41Mi/s
BM_validate_utf8_small/3      951668 ns       948250 ns          741 bytes_per_second=247.16Mi/s
BM_validate_utf8_small/4      944975 ns       942727 ns          743 bytes_per_second=331.473Mi/s
BM_validate_utf8_small/5      942938 ns       940954 ns          735 bytes_per_second=415.117Mi/s
BM_validate_utf8_small/6      841230 ns       839293 ns          833 bytes_per_second=558.472Mi/s
BM_validate_utf8_small/7      891388 ns       889915 ns          786 bytes_per_second=614.48Mi/s
BM_validate_utf8_small/8      663682 ns       662348 ns         1054 bytes_per_second=943.532Mi/s
BM_validate_utf8_small/9      841233 ns       839826 ns          834 bytes_per_second=837.145Mi/s
BM_validate_utf8_small/10     849720 ns       846173 ns          833 bytes_per_second=923.173Mi/s
[...]

PR #986, opened 2026-06-03
perf: short-input scalar fallback in convert_utf8_to_utf16{le,be} span overloads

aryanputta:
## What changed

This adds a `<= 16` byte scalar fallback to the UTF-8 to UTF-16 span overloads:

- `convert_utf8_to_utf16le(std::span<const char>, std::span<char16_t>)`
- `convert_utf8_to_utf16be(std::span<const char>, std::span<char16_t>)`
- `convert_utf8_to_utf16le_with_errors(std::span<const char>, std::span<char16_t>)`
- `convert_utf8_to_utf16be_with_errors(std::span<const char>, std::span<char16_t>)`

For those short inputs, the span overload calls the scalar implementation directly instead of going through the selected implementation dispatch path.

## Scope

This keeps the optimization on the inline span/ranges path.

It does not change the C-style API, does not add a new public inline API, and does not move the fix into `implementation.cpp`. That keeps the patch limited to the non-breaking path discussed in issue `#925`.

## Benchmark data

Target reported by `shortbench`: `haswell`

```text
Function                                      Size   Pointer   Span
convert_utf8_to_utf16le                         1    10.9 ns   4.4 ns
convert_utf8_to_utf16le                        16    21.4 ns   6.7 ns
convert_utf8_to_utf16le                        31    34.9 ns   36.7 ns

convert_utf8_to_utf16le_with_errors             1    11.4 ns   3.6 ns
convert_utf8_to_utf16le_with_errors            16    19.7 ns   5.9 ns
```

The size `31` result is why the cutoff stays conservative. The scalar shortcut helps where dispatch overhead dominates, then the existing selected implementation path becomes competitive again.

`shortbench` now has pointer and span entries for the normal and `_with_errors` UTF-8 to UTF-16 paths, so each optimized span overload has a direct C-style baseline.

## Tests

The span tests compare the optimized overloads against the existing pointer API for:

- valid ASCII inputs
- multibyte UTF-8 inputs
- threshold boundary sizes `14, 15, 16, 17, 18`
- invalid UTF-8 cases: illegal bytes, truncated sequence, overlong encoding, surrogate encoding, and out-of-range code point

Local validation:

```text
ctest --output-on-failure -R '^span_tests$'
100% tests passed, 0 tests failed out of 1

ctest --output-on-failure -j 4
100% tests passed, 0 tests failed out of 91
```

### DIFF
--- a/benchmarks/shortbench.cpp
+++ b/benchmarks/shortbench.cpp
@@ -167,6 +167,18 @@ std::vector<BenchmarkFunc> available_functions = {
          return len;
        };
      }},
+    // Span-overload variant: calls the simdutf_really_inline span path in
+    // implementation.h which, for short inputs (<=16 bytes), calls the scalar
+    // implementation directly without going through the SIMD dispatcher.
+    // Compare against convert_utf8_to_utf16le to see the short-input benefit.
+    {"convert_utf8_to_utf16le_span",
+     [](std::span<const char> input, std::span<char> output) {
+       return [input, output]() -> size_t {
+         std::span<char16_t> out16{reinterpret_cast<char16_t *>(output.data()),
+                                   output.size() / sizeof(char16_t)};
+         return simdutf::convert_utf8_to_utf16le(input, out16);
+       };
+     }},
     {"convert_utf8_to_utf16be",
      [](std::span<const char> input, std::span<char> output) {
        return [input, output]() -> size_t {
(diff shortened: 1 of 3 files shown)


aryanputta:
Revised after studying #925 thread, #926, and #957.

**Previous version of this PR was wrong.** I had added the short-input check to `implementation.cpp`, which is exactly what PR #926 attempted. @lemire correctly identified the problem: the function call is still not inlineable, so the branch "buys you little and might even cost you a bit."

Tested it explicitly: an `implementation.cpp`-only change shows **0% improvement** in shortbench. Confirmed.

**The fix: span overloads in `implementation.h`**

The span overloads are `simdutf_really_inline`. Since `implementation.h` already includes the scalar utf8_to_utf16 headers at lines 166–167, the scalar `convert<>()` is visible and inlineable from the inline body. Adding the threshold check there lets the compiler eliminate both the function call frame and the vtable dispatch for short strings.

**Hard numbers** (Apple M-series / haswell emulation, pre-allocated buffers, no heap in loop, 5M reps):

```
Input size    C-style (unchanged)   span (this PR)   ratio
------------------------------------------------------------------
  2 bytes          13.88 ns              5.83 ns      2.38x FASTER
 16 bytes          23.14 ns              5.58 ns      4.15x FASTER
 63 bytes          38.59 ns             37.24 ns      1.04x same
255 bytes          42.37 ns             39.86 ns      1.06x same
```

Scalar path confirmed being called: `convert_utf8_to_utf16le(span{"\xFF\xFF", 2}, ...)` returns 0 (invalid UTF-8 rejected), matching the scalar reference.

**Scope**: this only helps C++ callers using `std::span`. The C-style `(const char*, size_t)` API is untouched — that fix requires the ABI-breaking inline promotion discussed in the thread. `src/implementation.cpp` has **no changes** in this version.

Happy to run on other platforms if that would help.

aryanputta:
Updated 

**What changed in this push:**

**`benchmarks/shortbench.cpp`** — four new entries that exercise the exact changed call path:
- `convert_utf8_to_utf16le_span`
- `convert_utf8_to_utf16be_span`
- `convert_utf8_to_utf16le_with_errors_span`
- `convert_utf8_to_utf16be_with_errors_span`

These call `simdutf::convert_utf8_to_utf16le(std::span<const char>{…}, std::span<char16_t>{…})` directly, hitting the `simdutf_really_inline` overload in `implementation.h`. You can now compare the C-style and span variants side by side:

```bash
cmake -B build -DSIMDUTF_BENCHMARKS=ON -DSIMDUTF_TESTS=ON -DSIMDUTF_CXX_STANDARD=20
cmake --build build --target shortbench

./build/benchmarks/shortbench --function convert_utf8_to_utf16le       # C-style baseline
./build/benchmarks/shortbench --function convert_utf8_to_utf16le_span  # span fast path
```

Results on this machine (Apple / haswell emulation):

```
Size    C-style    span      ratio
1       9.5 ns     4.2 ns    2.26x FASTER
11      28.9 ns    12.7 ns   2.28x FASTER
21      23.3 ns    24.8 ns   same
31      39.9 ns    41.4 ns   same
```

Above 16 bytes both paths behave identically (expected — the threshold check is false and the span overload falls through to the same C-style function).

**`tests/span_tests.cpp`** — 11 new TEST() entries added to the existing span test file:
- ASCII, 2-byte (é), 3-byte (漢), 4-byte (😀) sequences, span output ≡ C-style reference
- Threshold boundary: sizes 14, 15, 16 (scalar branch) and 17, 18 (dispatch)
- Exactly 16 bytes of 2-byte sequences confirms scalar branch
- `_with_errors` variants agree on error code and position for bad continuation byte
- LE vs BE byte order verified distinct for non-ASCII output

```bash
cmake -B build -DSIMDUTF_CXX_STANDARD=20
cmake --build build --target span_tests
./build/tests/span_tests
# All 11 new tests: OK
```

`src/implementation.cpp` still has no changes.

lemire:
Running tests. This should be investigated.

PR #997, opened 2026-07-10
using __builtin_popcountll in the hope that it will pick up better optimizations

lemire:
On recent ARM processors, we can do better than NEON for popcnt.


clausecker:
One day, we can just use `stdc_count_ones()` from `<cstdbit>`.

lemire:
@clausecker True. Or the equivalent in C++20.

PR #1000, opened 2026-07-17
arm64/base64: reuse scalar base64 tables for encoding

tantei3:
Currently the routine has additional 64 byte tables for base64 encoding and performs the loading using LD4 which does additional de-interleaving. We can reuse the existing scalar tables and use LD1 which will load the registers without extra de-interleaving. LD1 is simpler and slightly faster than LD4 on various ARM cores.


lemire:
Good

PR #1001, opened 2026-07-20
arm64/base64: optimize insert_line_feed16

tantei3:
Since last element in input vector is not used, replace it with '\n' and use 15 instead of 0x80 as shuffle index to obtain newline instead of using extra compare and blend operations.


lemire:
I love this clever hack.

PR #1003, opened 2026-07-20
adding optimization from PR 1001 to westmere and haswell

lemire:
Direct follow up from https://github.com/simdutf/simdutf/pull/1001 by @tantei3


tantei3:
Westmere code looks good, however for Haswell I had a slightly different approach.
```cpp
simdutf_really_inline __m256i insert_line_feed32(__m256i input, int K) {
  __m128i lo = _mm256_extracti128_si256(input, 0);
  __m128i hi = _mm256_extracti128_si256(input, 1);
  if (K >= 16) {
    // lo is unchanged, hi needs to have newline inserted in between
    hi = insert_line_feed16(hi, K - 16);
  } else {
      // shift input by 1 byte, specifically lo's 15th byte into hi
      hi = _mm_alignr_epi8(hi, lo, 15);
      // lo needs to have newline inserted in between
      lo = insert_line_feed16(lo, K);
  }

  // Combine the lo and hi vectors
  __m256i result = _mm256_castsi128_si256(lo);
  result = _mm256_inserti128_si256(result, hi, 1);
  return result;
}
```
I split the __m256i into two __m128i chunks and depending on the K value changed the hi or lo part. I think the instruction counts are identical with the current diff, but the main advantage here is that the lookup table used is 256 byte instead of 1024, and the code is potentially reused with Westmere's implementation.

PR #1006, opened 2026-07-21
potential fix for vs issue

lemire:
There might be a bug in VS.


lemire:
@pauldreik There is a bug in visual studio.

Reproduction steps:

Start with a SIMD register x containing packed 16-bit signed integers: 4, 4, 4, ...

Create a second SIMD register m filled with all bits set to 1 (i.e. every 16-bit lane is 0xFFFF, which is -1 in two's complement).

Add m to x twice using signed 16-bit addition.

Expected result: Adding -1 twice is equivalent to subtracting 2, so the final values should be 2, 2, 2, ...

What actually happens: Visual Studio's compiler incorrectly precomputes m + m and treats the result as 32-bit signed integers (-2, -2, ...). This produces a completely wrong final result.

Godbolt: https://godbolt.org/z/8noM3Ka98

pauldreik:
interesting!

lemire:
@pauldreik Someone reported it formally. (A Microsoft Principal Engineer, I think.)

PR #1007, opened 2026-07-23
include what you use - std::abort

pauldreik:
this adds include of cstdlib everywhere std::abort is used.

The first commit is where it is absolutely needed.

The second commit is where is is strictly not needed, because it is transitively included.

Best practice is to include what you use. If you use std::abort, then include cstdlib.

The background to this is that I am working with an experimental standard library which has less transitive includes than libstdc++ and libc++.

### DIFF
--- a/benchmarks/base64/openssl3_base64.h
+++ b/benchmarks/base64/openssl3_base64.h
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdio>
+#include <cstdlib>
 namespace openssl3 {
 struct evp_Encode_Ctx_st {
   /* number saved in a partial encode/decode */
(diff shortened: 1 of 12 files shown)


lemire (approved):
Yeah so this can't possibly cause much harm.

pauldreik:
@lemire  I have another thing I am considering as a potential improvement. for the core library (excluding benchmarks and tests) we include algorithm just to get std::min, and possibly std::distance on pointers. we could implement those two . considering algorithm is pretty heavy, it could be an easy win. what do you think?

lemire:
@pauldreik Sure. I think we sort of discuss this with @mitchellh a while back (reducing our dependency on the standard lib). I think that min/max and distance are just silly and anyone can roll their own.

pauldreik:
> @pauldreik Sure. I think we sort of discuss this with @mitchellh a while back (reducing our dependency on the standard lib). I think that min/max and distance are just silly and anyone can roll their own.

I went ahead and made https://github.com/simdutf/simdutf/pull/1010

PR #1012, opened 2026-07-27
icelake/base64: optimized newline encoding for line_len < 64

tantei3:
Split the __m512i into two __m256i and expand the vector to have newlines at multiple of line_len bytes shifted by offset.
This avoids writing the vector to stack and using a scalar loop to encode the newlines.


lemire:
This makes a lot of sense to me.

PR #1014, opened 2026-07-29
aarch64/base64: Add NEON path for line_length < 16 encoding

tantei3:
Split the 16 byte vector into two 8 byte vectors and reuse the insert_line_feed16 method to add one or two newlines and write the two vectors to achieve the operation instead of writing to stack and using scalar loop.

### DIFF
--- a/src/arm64/arm_base64.cpp
+++ b/src/arm64/arm_base64.cpp
@@ -29,7 +29,7 @@
 /**
  * Insert a line feed character in the 16-byte input at index K in [0,16).
  */
-inline uint8x16_t insert_line_feed16(uint8x16_t input, size_t K) {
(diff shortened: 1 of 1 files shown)


lemire:
@tantei3 This adds a lot of code, for a case that is a bit extreme (people dont have such short lines typically).

Adding so much code can affect the code generation and makes the common cases slower.

tantei3:
Seems like the function assembly code size increases by about 100 instructions which makes the function non-inline. I guess it may not be a worthwhile trade-off for the common case, therefore I'm closing this PR. 
Anyways I had fun implementing this algorithm!

lemire:
@tantei3 We could have supported it with a cold path, but I am not sure it is an important use case.

lemire:
I agree that your code is fun.

PR #1015, opened 2026-07-31
icelake/base64: Add SIMD tail for binary_length_from_base64

tantei3:
Replaced the scalar loop with masked AVX512 instructions.

Studied the code generated https://gcc.godbolt.org/z/YGWWrzaE6 and observed shorter more optimum code instead of compiler's autovectorized part which looks long and bloated.


lemire:
Merged. I don't expect benefits, but it is nice.

PR #1016, opened 2026-07-31
Draft: investigate safe UTF-16 ASCII preflight

perfloop-agent:
## Status

Draft evidence handoff only. This PR is not ready to merge or share as a general performance result.

## What changed

This branch contains the candidate optimization, its focused benchmark and correctness tests, plus doc/safe_utf16_to_utf8_preflight.md. The document records the evidence that is supported and the proof that is still missing.

## Supported result

The public safe UTF-16-to-UTF-8 API has a sealed paired win for Haswell ASCII at N=256 and exact capacity: 139.904 to 42.504 ns/op, a 97.488 ns/op paired-median improvement across ten alternating pairs.

## Blocking residual

At N=256, ASCII except U+00E9 at index 127 leaves sample positions 0, 128, and 255 ASCII. The candidate enters the vector ASCII preflight, discovers the interior non-ASCII code unit, and then falls back to the old path after extra work. Existing benchmark modes never cover that branch.

The committed handoff names the required sealed public-API paired Haswell selector, fallback/guard checks, and explicit no-regression gate. Do not repair the optimization in this PR; assign that to the next session.

## Verification

- Targeted Release build: convert_utf16_to_utf8_safe_tests and benchmark_safe_utf16_to_utf8 passed locally.
- Artifact chain: the candidate Git bundle and patch hashes match the recorded candidate; candidate 38d0fc6 is directly atop comparison/harness commit af4bf61.

## AI disclosure

This is an AI-assisted draft created under operator direction. It remains draft pending human review of the evidence and wording before it is offered for maintainer review.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -98,4 +98,9 @@ add_executable(shortbench shortbench.cpp)
 target_link_libraries(shortbench PUBLIC simdutf::benchmarks::benchmark)
 target_compile_features(shortbench PRIVATE cxx_std_20)
 
+add_executable(benchmark_safe_utf16_to_utf8 benchmark_safe_utf16_to_utf8.cpp)
+target_link_libraries(benchmark_safe_utf16_to_utf8 PRIVATE simdutf)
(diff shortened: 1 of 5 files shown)


tsenart:
Closed at operator request; residual handoff withdrawn.

PR #1017, opened 2026-08-04
Add single-pass Latin-1 UTF-8 string materialization

perfloop-agent:
## Summary

Materializing Latin-1 into a `std::string` today takes two passes: `utf8_length_from_latin1` scans the whole input to size the output exactly, then the conversion scans it again. This PR adds a C++23 resizable-string overload of `simdutf::convert_latin1_to_utf8` that does it in one pass:

- checks `length > output.max_size() / 2` before forming the `length * 2` bound;
- `resize_and_overwrite` to that checked 2× bound, one dispatched conversion, then commits the exact written size;
- the exact-size pointer APIs, `_safe` variants, and constexpr paths are unchanged — callers that need exact-size allocation keep the existing preflight.

The deliberate tradeoff: the new path allocates a 2× bound instead of the exact size (65.6 KB → 131.1 KB per op on 64 KiB ASCII input). Allocation count is unchanged (1/op) and measured peak RSS did not regress.

## Performance

Adjacent before/after builds, fresh processes (recorded verification runs, x86-64):

| workload | before | after |
|---|---|---|
| 64 KiB ASCII, cold output | ~3830 ns/op | ~3067 ns/op |
| 64 KiB 50% high-bit, cold | ~15134 ns/op | ~14475 ns/op |
| 1 MiB ASCII, reused output | ~70.4 µs/op | ~50.4 µs/op |
| 1 MiB high-bit, reused | ~267 µs/op | ~240 µs/op |

Independently reproduced on Apple Silicon (clang, arm64): 64 KiB ASCII cold ~3812 → ~1522 ns/op; 50% high-bit ~11.1 → ~8.7 µs/op.

## Testing

- All 16 latin1 conversion test suites pass, including new tests for the resizable overload (cutoff boundary, reuse, custom-allocator `max_size` rejection, every input length 0–299 with varied bytes).
- C++17 and C++20 compatibility builds pass; the overload is gated on `resize_and_overwrite` feature detection.
- The safe-conversion fuzzer is extended to the new overload; amalgamation and single-header builds pass.

Full verification record: https://app.perfloop.ai/t/oss/case_ch8m0vvpwv

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@
   - [Base64](#base64)
   - [Find](#find)
   - [C++20 and std::span usage in simdutf](#c20-and-stdspan-usage-in-simdutf)
+  - [C++23 resizable Latin1-to-UTF-8 output](#c23-resizable-latin1-to-utf-8-output)
   - [C++23 and constexpr support](#c23-and-constexpr-support)
   - [Command-line tools](#command-line-tools)
   - [Manual implementation selection](#manual-implementation-selection)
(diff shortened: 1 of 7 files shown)


perfloop-agent:
Withdrawing this. On further review it doesn't hold up as a general improvement.

The resizable-output overload allocates — it reserves the worst-case 2 bytes per Latin-1 byte up front via `resize_and_overwrite` and never shrinks — and it introduces an allocation/throw path. That runs against simdutf's exception-free, non-allocating design (this change even had to narrow that guarantee in the README). It also regresses short-string latency, and on cold ASCII inputs the transient allocation doubles. The speedup comes from skipping the separate sizing pass, so the extra allocation and the win can't be decoupled.

The gain is real only for large-input, buffer-reuse workloads. If that's worth pursuing, it belongs as an explicitly opt-in resizable-output overload with the memory tradeoff documented — not as a general change to the library's guarantees. Apologies for the noise.

